PR #42, opened 2015-02-01
Add method to VisitDomainWrapper to get only unique diagnoses

mseaton:
@djazayeri , can you have a look at this?   The boolean parameters probably aren't needed, but I wasn't sure if we were intentionally excluding secondary diagnoses on all of the places we want to use this, so I added it in for more flexibility.  We can remove it if you don't like it...


rkorytkowski:
Pinging @djazayeri again. 

Also is there any issue in JIRA for this?

mseaton:
@rkorytkowski the ticket is EA-38 .  @djazayeri  last chance to object :)

djazayeri:
Seems fine to me.

PR #44, opened 2015-03-04
EA-45,49,50 Added validation for patient update, Modified getConditionHi...

sravanthi17:
...story base on the feedback, also changed getActiveconditions to return in reverse chronological order

### DIFF
--- a/condition-list/src/main/java/org/openmrs/api/db/hibernate/HibernateConditionDAO.java
+++ b/condition-list/src/main/java/org/openmrs/api/db/hibernate/HibernateConditionDAO.java
@@ -59,7 +59,8 @@ public class HibernateConditionDAO implements ConditionDAO {
     @Override
     @Transactional(readOnly = true)
     public List<Condition> getConditionHistory(Patient patient) {
-        Query query = sessionFactory.getCurrentSession().createQuery("select con from Condition as con inner join fetch con.concept as c where con.patient.patientId = :patientId and con.voided = false  group by c.conceptId, con.conditionId order by con.dateCreated desc");
+        Query query = sessionFactory.getCurrentSession().createQuery("select con from Condition as con where con.patient.patientId = :patientId and con.voided = false " +
+                "order by con.dateCreated desc");
         query.setInteger("patientId", patient.getId());
         return query.list();
     }
--- a/condition-list/src/main/java/org/openmrs/api/db/hibernate/HibernateConditionDAO.java
+++ b/condition-list/src/main/java/org/openmrs/api/db/hibernate/HibernateConditionDAO.java
@@ -59,9 +59,7 @@ public class HibernateConditionDAO implements ConditionDAO {
     @Override
     @Transactional(readOnly = true)
     public List<Condition> getConditionHistory(Patient patient) {
-        Query query = sessionFactory.getCurrentSession().createQuery("select con from Condition as con " +
-                " inner join fetch con.concept as c left join c.names as name with name.conceptNameType = 'FULLY_SPECIFIED' " +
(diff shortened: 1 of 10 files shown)


condition-list/src/main/java/org/openmrs/api/impl/ConditionServiceImpl.java
```diff
@@ -49,8 +52,46 @@ public Condition save(Condition condition) {
...
+
+        List<Condition> groupedConditions = new ArrayList<Condition>();
+
+        if (CollectionUtils.isEmpty(conditionList)) return new ArrayList<ConditionHistory>();
+
+        ConditionHistory conditionhistory = createConditionHistory(conditionList.get(0));
+        groupedConditions.add(conditionList.get(0));
+        for (int i = 1; i < conditionList.size(); i++) {
+            Condition condition = conditionList.get(i);
+            if (condition.getConcept().equals(conditionhistory.getCondition())) {
```
rkorytkowski: This assumes conditionList will have conditions ordered by concepts, which I don't think is true. They are ordered by dateCreated. I think what you need here is Map<Concept, ConditionHistory> and Map<String, ConditionHistory> (for non coded conditions) and then you can distribute conditionList based on concept or nonCoded value into proper ConditionHistories.

condition-list/src/main/java/org/openmrs/api/db/hibernate/HibernateConditionDAO.java
```diff
@@ -59,9 +59,7 @@ public Condition getConditionByUuid(String uuid) {
     @Override
     @Transactional(readOnly = true)
     public List<Condition> getConditionHistory(Patient patient) {
-        Query query = sessionFactory.getCurrentSession().createQuery("select con from Condition as con " +
-                " inner join fetch con.concept as c left join c.names as name with name.conceptNameType = 'FULLY_SPECIFIED' " +
-                " where con.patient.patientId = :patientId and con.voided = false  group by c.conceptId, con.conditionId order by name.name asc , con.dateCreated desc");
+        Query query = sessionFactory.getCurrentSession().createQuery("select con from Condition as con inner join fetch con.concept as c where con.patient.patientId = :patientId and con.voided = false  group by c.conceptId, con.conditionId order by con.dateCreated desc");
```
rkorytkowski: Can you explain the purpose of "group by" please? I don't understand why you use it here.

condition-list/src/main/java/org/openmrs/api/validator/ConditionValidator.java
```diff
@@ -121,4 +122,11 @@ private void validateNonCodedCondition(Condition condition, Errors errors) {
             }
         }
     }
+
+    private void validatePatient(Condition condition, Errors errors) {
+        Condition existingCondition = conditionService.getConditionByUuid(condition.getUuid());
```
rkorytkowski: Can you please instead modify Condition.setPatient to throw an exception if you try to change the value? Please see my last comment on https://issues.openmrs.org/browse/EA-45

rkorytkowski:
Thanks @sravanthi17 !

sravanthi17:
Welcome.. :)

On Fri, Mar 20, 2015 at 6:33 PM, Rafal Korytkowski <notifications@github.com
> wrote:

> Thanks @sravanthi17 <https://github.com/sravanthi17> !
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/openmrs/openmrs-module-emrapi/pull/44#issuecomment-84007172>
> .
>



-- 
*N. SRAVANTHI,*
*Application Developer,*
*ThoughtWorks,*

PR #46, opened 2015-03-11
Adding emr.extraPatientIdentifierTypes Global Property

craigappl:
This global property wasn't migrated from the PIH/openmrs-module-emr, but is used in the EMRAPI getextraidentifiers method. It's also used in the coreapps patient header to display multiple identifiers.

The discussion is available in the dev forum https://groups.google.com/a/openmrs.org/forum/#!topic/dev/s6M719M6loM


mogoodrich:
Thanks Craig!

PR #48, opened 2015-04-01
EA-61: Create german messages.properties

STheis:
German translation of emr api


djazayeri:
This module isn't in transifex (yet), so I'm going to merge this here.

mogoodrich:
Ah right, duh, thanks @djazayeri

PR #52, opened 2015-05-14
EA-65: Fixed the bug that the OrderType is missing in TestOrder

SwathiVarkala:
OrderType was missing in TestOrder. Fixed the same.


vinayvenu:
OrderType is made mandatory in OrderServiceImpl as of 1.10, and the field has been made mandatory later. It is not necessary to make a null check when retrieving OrderType.

PR #54, opened 2015-05-29
EA-68 - Handling the Hibernate lazy loading problem while querying th…

bharatak:
When trying to search for OpenMRS TestOrder using uuid, the returned Order object is a proxy because of Hibernate's lazy loading. When the Order is casted to TestOrder, it throws a ClassCastException.  The fix is to load the TestOrder properly (using HibernateLazyLoader) before casting it.


bharatak:
@dkayiwa Sorry for missing it out.  I have fixed it now.

PR #64, opened 2015-07-25
EA-75 : Fix the datemapper issue - remove setting default date as 01-…

preethi29:
…01-1970


preethi29:
@vinayvenu Have checked the diagnosis service. It does not make much difference if we pass null. If null is passed, no fromDate criteria is added and if new Date(0) is added, 1970-01-01 is added as fromDate criteria.

PR #84, opened 2016-01-22
EA-90: EncounterTransaction should support specifying EncounterRole f…

mogoodrich:
…or a Provider


bharatak:
Hi @mogoodrich I think it would be better if it is called encounterRoleUuid.  It will be more clear looking at the member variable what it refers to.  The similar approach is used for encounterUuid, visitUuid, locationUuid etc.  Otherwise, everything looks good.

mogoodrich:
Makes sense to me re:encounterRoleUuid.  I have made this change.

PR #91, opened 2016-04-06
EA-95 Add missing global property emr.encounterMatcher to config.xml

bgeVam:
see https://issues.openmrs.org/browse/EA-95

### DIFF
--- a/omod/src/main/resources/config.xml
+++ b/omod/src/main/resources/config.xml
@@ -168,6 +168,14 @@
             UUID of the Discharge Form (not required)
         </description>
     </globalProperty>
+    
+    <globalProperty>
+        <property>emr.encounterMatcher</property>
+        <defaultValue></defaultValue>
+        <description>
+            Class name of a custom encounter matcher (e.g. org.openmrs.module.example.customEncounterMatcher)
+        </description>
+    </globalProperty>
 
     <globalProperty>
         <property>emrapi.suppressedDiagnosisConcepts</property>


omod/src/main/resources/config.xml
```diff
@@ -168,6 +168,14 @@
             UUID of the Discharge Form (not required)
         </description>
     </globalProperty>
+    
+    <globalProperty>
+        <property>emr.encounterMatcher</property>
+        <defaultValue></defaultValue>
+        <description>
+            Custom Encounter Matcher
```
dkayiwa: Does it take a class name? Or something else? Can you give an example in the description?

bgeVam: PR updated, thank you for your feedback

omod/src/main/resources/config.xml
```diff
@@ -168,6 +168,14 @@
             UUID of the Discharge Form (not required)
         </description>
     </globalProperty>
+    
+    <globalProperty>
+        <property>emr.encounterMatcher</property>
+        <defaultValue></defaultValue>
+        <description>
+            Class name of a custom encounter matcher (e.g. org.openmrs.module.example.customEncounterMatcher)
```
dkayiwa: The convention is to have class names that begin with capital letters instead of small. c vs C :)

omod/src/main/resources/config.xml
```diff
@@ -168,6 +168,14 @@
             UUID of the Discharge Form (not required)
         </description>
     </globalProperty>
+    
+    <globalProperty>
+        <property>emr.encounterMatcher</property>
+        <defaultValue></defaultValue>
```
dkayiwa: By the way, does it have a default value? The one used by openmrs if none is specified. If yes, then it can be here.

bgeVam: The current logic implied is to take the DefaultEncounterMatcher if no EncounterMatcher is specified, I would not change this, in my opinion the aim of the field is just to set it if a custom matcher is specified.
If we would set DefaultEncounterMatcher by default, the check in the logic would be unnecessary

PR #93, opened 2016-06-12
Correct fallback of concept names in a locale

aj-jaswanth:
1) If short name is not available for a concept in a locale use its fully specified name in that locale
2) If fully specified name is also not available in that locale, use short name name of that concept in the system default locale
3) If short name is not available for the system default locale, use its fully specified name

### DIFF
--- a/api/src/main/java/org/openmrs/module/emrapi/encounter/ConceptMapper.java
+++ b/api/src/main/java/org/openmrs/module/emrapi/encounter/ConceptMapper.java
@@ -24,9 +24,11 @@ import org.openmrs.module.emrapi.utils.HibernateLazyLoader;
 import org.openmrs.util.LocaleUtility;
 import org.openmrs.util.OpenmrsConstants;
 
+import java.util.Locale;
+
 public class ConceptMapper {
     public EncounterTransaction.Concept map(Concept concept) {
-        if (concept == null){
+        if (concept == null) {
             return null;
         }
         concept = new HibernateLazyLoader().load(concept);
@@ -42,7 +44,8 @@ public class ConceptMapper {
                 conceptClassName,
                 getShortName(concept),
                 concept.getConceptMappings());
-        if(concept.isNumeric()) {
+
+        if (concept.isNumeric()) {
             encounterTransactionConcept.setUnits(((ConceptNumeric) concept).getUnits());
             encounterTransactionConcept.setHiNormal(((ConceptNumeric) concept).getHiNormal());
             encounterTransactionConcept.setLowNormal(((ConceptNumeric) concept).getLowNormal());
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/openmrs/module/emrapi/encounter/ConceptMapper.java
```diff
@@ -54,16 +56,24 @@ private String getShortName(Concept concept) {
...
+            Locale locale = LocaleUtility.fromSpecification(userDefaultLocale);
+            ConceptName shortNameInLocale = concept.getShortNameInLocale(locale);
             if(shortNameInLocale != null) {
                 shortName = shortNameInLocale.getName();
+            } else {
+                ConceptName fullySpecifiedName = concept.getFullySpecifiedName(locale);
+                if (fullySpecifiedName != null) {
+                    shortName = fullySpecifiedName.getName();
+                } else {
+                    String systemDefaultLocale = Context.getAdministrationService().getGlobalProperty("default_locale");
```
dkayiwa: OpenmrsConstants.GLOBAL_PROPERTY_DEFAULT_LOCALE would be better than "default_locale" :)

api/src/main/java/org/openmrs/module/emrapi/encounter/ConceptMapper.java
```diff
@@ -54,16 +56,24 @@ private String getShortName(Concept concept) {
...
+                    if (systemDefaultLocale != null) {
+                        Locale defLocale = LocaleUtility.fromSpecification(systemDefaultLocale);
+                        shortName = concept.getFullySpecifiedName(defLocale).getName();
+                    }
+                }
             }
         }
-        if(shortName == null) {
-            shortName = concept.getShortNames() != null && concept.getShortNames().size() > 0 ? concept.getShortNames().iterator().next().getName() : null;
```
dkayiwa: If we reach here when shortName is still null, wouldn't it be better to leave this code as it was than just returning null?

aj-jaswanth:
@dkayiwa Changed the fall back to use short name in the system default locale also.
Regarding returning null if all the fallbacks fail, I was thinking that it enforces the user to atleast specify fully specified names for the default locale.

dkayiwa:
@aj-jaswanth i have seen a few user dictionaries with some concepts not having names in the default locale. Do you want such to end up with null pointer exceptions? Or you would rather have some name in some locale as it has always been?

aj-jaswanth:
@dkayiwa Makes sense. Added code to handle the situation.

api/src/main/java/org/openmrs/module/emrapi/encounter/ConceptMapper.java
```diff
@@ -53,17 +56,43 @@
...
+            shortName = getAnAvailableName(concept, userDefaultLocale);
+            if (shortName == null) {
+                Locale systemDefaultLocale = getSystemDefaultLocale();
+                shortName = getAnAvailableName(concept, systemDefaultLocale);
             }
         }
-        if(shortName == null) {
-            shortName = concept.getShortNames() != null && concept.getShortNames().size() > 0 ? concept.getShortNames().iterator().next().getName() : null;
+        if (shortName == null) {
+            shortName = concept.getName().getName();
```
dkayiwa: Did you intentionally change from concept.getShortNames() to concept.getName()

aj-jaswanth: Yes. concept.getName() is using locale.allowed.list to find locales. It is better than picking up a name randomly. It tries to pickup a preferred name.

PR #95, opened 2016-07-19
EA-101  Considering visit location while closing stale visits

hanisha93:
As part of 'Close Stale Visits Task' we are closing the following visits
1. which do not have location tagged to it.(location_id is null)
2. which have visit location tagged to it.

### DIFF
--- a/api/src/main/java/org/openmrs/module/emrapi/adt/AdtServiceImpl.java
+++ b/api/src/main/java/org/openmrs/module/emrapi/adt/AdtServiceImpl.java
@@ -28,6 +28,7 @@ import org.openmrs.PersonAttributeType;
 import org.openmrs.Provider;
 import org.openmrs.User;
 import org.openmrs.Visit;
+import org.openmrs.LocationTag;
 import org.openmrs.api.APIException;
 import org.openmrs.api.EncounterService;
 import org.openmrs.api.LocationService;
@@ -43,6 +44,7 @@ import org.openmrs.module.emrapi.disposition.Disposition;
 import org.openmrs.module.emrapi.domainwrapper.DomainWrapperFactory;
 import org.openmrs.module.emrapi.merge.PatientMergeAction;
 import org.openmrs.module.emrapi.patient.PatientDomainWrapper;
+import org.openmrs.module.emrapi.visit.EmrVisitService;
 import org.openmrs.module.emrapi.visit.VisitDomainWrapper;
 import org.openmrs.serialization.SerializationException;
 import org.openmrs.util.OpenmrsUtil;
@@ -60,6 +62,7 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.HashSet;
 
 
 public class AdtServiceImpl extends BaseOpenmrsService implements AdtService {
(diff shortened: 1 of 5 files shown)


djazayeri:
Brief comments from skimming this:

1. The first comment on this PR doesn't seem to match the logic you're implementing.
2. I don't think it makes general business sense to dual-purpose "Login Location" to drive this logic. (See https://issues.openmrs.org/browse/EA-101?focusedCommentId=230859&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-230859 )
3. You would need to javadoc why you're introducing a new getVisits method (instead of using the openmrs-core one)
4. The code would read a log cleaner if you just added another test within shouldBeClosed(Visit). Would it really be a huge performance hit to do it that way? (Are you expecting many more outside-this-facility visits vs owned-by-this-facility ones?)

vinayvenu:
1. It looks like this is to guard against null locations that probably would have been there prior to this change. 
We should instead migrate existing implementations to actually add locations to all visits as part of the implementation. This will also help reduce quite a bit of code that we have included here (EmrVisitDao). 
2. I agree - @angshu - what do you think? 
4. This could be a problem, since we are expecting outside visits to be significantly higher than visits owned by this facility.

angshu:
Mostly related to 2)
Are we saying that we find all open visits matching certain criteria (open since last 2 days), and check whether the location belongs to current location, and close if applicable?
The question still remain about identifying what the "current" location? Ones that are tagged with "visit location", "roots of hierarchy", "visit location in the hierarchy"?? 


Lets take this example:

- Big Hospital 1
---- Dept 1 (Visit location)
--------- OPD-D1 (Login location - hence encounter location)
---- Dept 2 (Visit Location)
--------- OPD-D2 (Login location - hence encounter location)
---- Emergency (Login location - hence encounter location)
- Big Hospital 2
- Small Hospital 1


How will we address this scenario?

mogoodrich:
I was originally confused why you were dual purposing Login Location as @djazayeri  mentioned, but from reading the follow-on comments, I this may be to handle legacy implementations that weren't tagging locations with the "supports visit" tag?  That is the pseudo-code is that all possible locations include:

1) Locations tagged with the "supports visit" tag
2) Top-level parent location of any location with the "login location" tag.

I agree that #2 seems a little convoluted, and the code would read a lot easier without it, but I don't think it breaks our implementation at all.  Our workflow/logic should currently assure that:

1) All visits have a location
2) All locations in 1) are tagged as "supports visits"

djazayeri:
> This could be a problem, since we are expecting outside visits to be significantly higher than visits owned by this facility

But will outside visits _that are still open_ be significantly higher? If so, okay, I accept the optimization (over cleaner code).

In the scenario @angshu drew, I would expect Emergency to also be tagged as a Visit Location (at least, I'm assuming a patient can start their visit there). Given that scenario, we could purely use "Visit Location" to decide whether to close the visit, i.e. we only close visits where visit.location.hasTag("Visit Location"). Basically this would mean we clarify the meaning of "Visit Location" in the emrapi module to indicate that visits for this location are managed on this server, i.e. we both create them and close them for that location. (This assumes that you don't tag Big Hospital 2 or Small Hospital 1 with that tag.)

If for some reason you don't want to mark Emergency as supporting visits (why?) then I'd suggest a new "Manage Visits For This Location" tag which you'd put on Big Hospital 1.

I agree with @mogoodrich 's statement that
> 2) seems a little convoluted, and the code would read a lot easier without it, but if it's the only way, I don't think it breaks our implementation at all. (Just make sure you document the logic in the code?)

Finally, a +1 for properly migrating existing data so that all visits have locations, and not adding additional code to special-case the visit.location == null scenario.

vinayvenu:
Summarizing, the visit close logic would be
1. Retrieve open visits that are tagged "Visit Location"
2. Run rules to check if they need to be closed (make sure there is no disposition that allows the visit to continue, and there have not been any modifications in the past x hours)
3. Close if required. 

As opposed to the previous logic
1. **Retrieve all open visits. **
2. Run rules to check if they need to be closed (make sure there is no disposition that allows the visit to continue, and there have not been any modifications in the past x hours)
3. Close if required. 

According to @mogoodrich , this will work since all visits have locations tagged as "Visit Location". In Bahmni, we will make this mandatory. 

I am going ahead and merging this PR optimistically, please let me know if you feel there is a problem.

mogoodrich:
Looks good to me, thanks!

angshu:
I thought, the visit.location is either a) for a given location - any location in the hierarchy thats tagged with "visit location"  b) for a given location - root if none in the hierarchy is tagged with "visit location". 

Going by the above logic, are we saying that all locations must be tagged with "visit locations", including the top of the hierarchy?

djazayeri:
From my perspective, we don't need to handle the case where no location is tagged with "visit location". (I.e. this is a data/config issue to ensure at least one visit location)

angshu:
@djazayeri: that does imply that every implementation must specific the visit location. If thats the expectation, then we wouldn't need to do anything.

PR #96, opened 2016-07-21
EA-102 | Fetch active visits based on location.

chethandeshpande:


bharatak:
@chethandeshpande The location parameter might not be available always.  The code is not backward compatible.  Need to handle the case when the location is not available.

dkayiwa:
@bharatak hopefully you saw this https://ci.openmrs.org/browse/EMRAPI-EMRAPI-JOB1-1622/commit/ :)

bharatak:
@dkayiwa I saw this.  The testcase expects the order of elements from a set which is wrong.  @chethandeshpande is working on the fix for it.

PR #99, opened 2016-08-09
Ea104

sravanthi17:
Please find the openmrs issue link @https://issues.openmrs.org/browse/EA-104

### DIFF
--- a/api-1.10/pom.xml
+++ b/api-1.10/pom.xml
@@ -29,6 +29,12 @@
             <artifactId>openmrs-api</artifactId>
             <version>${openMRSVersion}</version>
             <scope>provided</scope>
+            <exclusions>
+                  <exclusion>
+                      <groupId>javassist</groupId>
+                      <artifactId>javassist</artifactId>
+                  </exclusion>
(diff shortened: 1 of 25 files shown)


sravanthi17:
Will raise a new request. Sorry for the inconvenience

mogoodrich:
For what it's worth, looks like the EMR-API build is still broken, even after this revert.  May have something to do with the version of Java we are using to compile?

dkayiwa:
@mogoodrich does it also fail locally? I suspect it could be a side effect of @maurya's bumping the platform version to 2.0

mogoodrich:
It fails on PIH bamboo as well.  Will test locallly now.

Where did we upgrade the platform version?  In the reference app distro?  I think this would definitely cause problems, especially if 2.0 requires Java 8?  We will have to figure out a strategy for this.

dkayiwa:
Yes in the refapp distro. Let me revert it for now to see if it fixes.

mogoodrich:
I'll post to talk about it, will @ you...

PR #104, opened 2016-10-20
EA-107: fix for "Editing some fields on Obs is not allowed" Error because of hibernate auto flush due to other hibernate call.

Gaurav-Deshkar:


api/src/main/java/org/openmrs/module/emrapi/encounter/EmrEncounterServiceImpl.java
```diff
@@ -106,6 +111,23 @@ public void onStartup() {
 
     @Override
     public EncounterTransaction save(EncounterTransaction encounterTransaction) {
+
+        EncounterTransaction updatedEncounterTransaction = null;
+
+        DbSession session = sessionFactory.getCurrentSession();
+        FlushMode flushMode = session.getFlushMode();
+        session.setFlushMode(FlushMode.MANUAL);
```
dkayiwa: @Gaurav-Deshkar have hibernate in the service layer instead of the data access layer does not look sweet. :)
Can you point me to a failing unit test and i see if i can make some alternative suggestions?

Gaurav-Deshkar: @dkayiwa this issue is not reproducible in a unit test. we have multiple get calls in this service call that's why I have put it in the service layer. for context please refer to https://talk.openmrs.org/t/obs-save-throws-cannot-edit-some-fields-on-obs-due-to-autoflush-of-other-db-operations/8321

dkayiwa: Yes i have been following that talk thread. If this is not easy to reproduce in a unit test, do you have any pointers to steps i could take to reproduce?
In the worst case, if we fail to get a decent solution, i would rather we create new methods in the data access layer for this hibernate magic and then let the service layer call them.

Gaurav-Deshkar: @dkayiwa , following steps to reproduce the issue.
1. create two obs , #obs1 and #obs2, where #obs2 must be coded obs.
2. update the #obs1 value.
use emr api endpoint `/emrapi/encounter`.

we have setup data at [https://upgrade.mybahmni.org](https://upgrade.mybahmni.org) . we have already performed step1, where we have added "Examination Notes" (text obs) as #obs1 and "Chief Complaint" (coded obs) as #obs2.  Now for step2 , which is to edit #obs1 ,we are trying to edit "Examination Notes". please hit the (POST) endpoint to perform step2 `https://upgrade.mybahmni.org/openmrs/ws/rest/emrapi/encounter` 
with following payload 

```
{
  "visitUuid": "bbd550b4-aaf6-4eb2-94e0-d6131e0fc3fb",
  "locationUuid": "c1e42932-3f10-11e4-adec-0800271c1b75",
  "patientUuid": "598e6d04-8c0d-49ca-a5a1-86918d237d4a",
  "visitTypeUuid": "c22a5000-3f10-11e4-adec-0800271c1b75",
  "encounterTypeUuid": "81852aee-3f10-11e4-adec-0800271c1b75",
  "disposition": null,
  "observations": [
    {
      "uuid": "17874c86-fbf2-4997-a8b9-a4f94c4ec47a",
      "value": "examination notes edited",
      "comment": null,
      "voided": false,
      "voidReason": null,
      "concept": {
        "uuid": "c39a35e2-3f10-11e4-adec-0800271c1b75",
        "name": "Examination Notes",
        "dataType": "Text",
        "shortName": "Examination Notes",
        "conceptClass": "Misc",
        "hiNormal": null,
        "lowNormal": null,
        "set": false,
        "mappings": []
      },
      "orderUuid": null,
      "creator": {
        "uuid": "c1c21e11-3f10-11e4-adec-0800271c1b75",
        "personName": "Super Man"
      },
      "formNamespace": null,
      "formFieldPath": null
    },
    {
      "uuid": "b6de1b37-27f0-463b-9006-5cafae167365",
      "value": "1dfb0948-4e15-11e4-8a57-0800271c1b75",
      "comment": null,
      "voided": false,
      "voidReason": null,
      "concept": {
        "uuid": "c3959ab5-3f10-11e4-adec-0800271c1b75",
        "name": "Chief Complaint",
        "dataType": "Coded",
        "shortName": "Chief Complaint",
        "conceptClass": "Misc",
        "hiNormal": null,
        "lowNormal": null,
        "set": false,
        "mappings": []
      },
      "orderUuid": null,
      "creator": {
        "uuid": "c1c21e11-3f10-11e4-adec-0800271c1b75",
        "personName": "Super Man"
      },
      "formNamespace": null,
      "formFieldPath": null
    }
  ],
  "orders": [],
  "drugOrders": [],
  "diagnoses": [],
  "providers": [
    {
      "uuid": "c1c26908-3f10-11e4-adec-0800271c1b75",
      "name": "Super Man",
      "encounterRoleUuid": "a0b03050-c99b-11e0-9572-0800200c9a66"
    }
  ],
  "extensions": {
    "mdrtbSpecimen": []
  },
  "context": {}
}
```

PS: To reproduce on local please update all the "uuid" accordingly.

preethi29: The above issue is due to autoflush which happens at ConceptService.getConceptByUuid at 
https://github.com/openmrs/openmrs-module-emrapi/blob/master/api/src/main/java/org/openmrs/module/emrapi/encounter/mapper/ObsMapper.java#L97
But it is not just this call, once this call is fixed, the autoflush will happen at https://github.com/openmrs/openmrs-module-emrapi/blob/master/api/src/main/java/org/openmrs/module/emrapi/encounter/EncounterDispositionServiceHelper.java#L158 when we add disposition

And there are also hooks supported in encounterTransaction save using EncounterTransactionHandler,  so if the client creates their own custom handler, it might break in their handlers as well. So we found this place to be more appropriate to add the manual flush. 
Also this should not be the long term solution, we would need to fix it in openmrs in a proper way after 2.0.1. Once it is solutioned and fixed, this can be removed.

dkayiwa: @preethi29 can we put the hibernate calls in the data access layer and then the service layer just calls the new data access layer methods to take care of this?

preethi29: If we put it in DAO layer, we will have to put it in multiple DAO calls and changes should be made in openmrs-core. And as I said before, we wont know what all read-only methods would cause auto-flush since we support hooks for EncounterTransaction which means client can use any read-only methods in their code. I guess the best ideal solution is to fix it properly in openmrs-core using some way (neither manual flush nor evictFromSession). But I think it is not planned for openmrs 2.0.1

dkayiwa: @preethi29 i mean creating a new method in the DAO to change the flush mode. Then you can create a service layer method to call into it. Or a utility class. The pattern am looking at is like what we have in Context.evictFromSession, Context.flushSession, etc. This way, the service layer does not have hibernate specific classes. Does it make sense?

preethi29: cool got it. the method can be created in emrapi ryt?

Gaurav-Deshkar:
@dkayiwa I have updated the pull request

PR #106, opened 2016-11-04
EA-109 | Fix for duplication of disposition obs group on edit and add…

preethi29:
… profile for running tests against openmrs2.0

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -55,11 +55,6 @@
             <scope>test</scope>
         </dependency>
         
-        <dependency>
-            <groupId>org.openmrs.module</groupId>
-            <artifactId>allergyapi-api</artifactId>
-            <scope>provided</scope>
-        </dependency>
 
         <!-- End OpenMRS core -->
     </dependencies>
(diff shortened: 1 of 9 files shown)


dkayiwa:
@preethi29 did you notice the travis failure?

preethi29:
@dkayiwa It passes in my local. Is there a way to view the surefire-reports?

dkayiwa:
@preethi29 did you already see this? https://travis-ci.org/openmrs/openmrs-module-emrapi/builds/173206694

preethi29:
Yeah I have taken a look at it. It just says error in tests. I would need to see the surefire-reports to see the detail of the error.

preethi29:
@dkayiwa Build passed. It was failing to case sensitive in dataset filename. Now its green

dkayiwa:
@preethi29 oh great! :)

preethi29:
@dkayiwa Thanks :)

dkayiwa:
@preethi29 does this compile successfully for you on Java 7?

preethi29:
@dkayiwa I have added MRS2.0 profile to run tests against openmrs 2 which essentially uses BaseModuleContextSensitiveTest that is compiled in java 8. Thats why when running tests in jdk 7, it throws java.lang.UnsupportedClassVersionError.

 Have you faced this kind of problem before since there are many modules which run tests against mrs 1.x and 2.x? If so, how was it solved?

dkayiwa:
@preethi29 as you can see here https://github.com/openmrs/openmrs-module-reporting/blob/master/pom.xml#L322 we do not make that profile active by default. We let whoever is compiling to activate that profile if they are using java 8

PR #115, opened 2016-11-30
EA-114: getUniqueDiagnoses(true, true) reimplemented with HQL.

mks-d:
The 4th and final case reimplemented ;-)

### DIFF
--- a/api/src/main/java/org/openmrs/module/emrapi/visit/VisitDomainWrapper.java
+++ b/api/src/main/java/org/openmrs/module/emrapi/visit/VisitDomainWrapper.java
@@ -344,9 +344,6 @@ public class VisitDomainWrapper implements DomainWrapper {
      */
     public List<Diagnosis> getUniqueDiagnoses(Boolean primaryOnly, Boolean confirmedOnly) {
        
-       if (primaryOnly == true && confirmedOnly == true) // TODO: Re-implement that specific case with HQL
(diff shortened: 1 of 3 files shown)


api/src/main/java/org/openmrs/module/emrapi/visit/VisitDomainWrapper.java
```diff
@@ -344,9 +344,6 @@ public Disposition getMostRecentDisposition() {
      */
     public List<Diagnosis> getUniqueDiagnoses(Boolean primaryOnly, Boolean confirmedOnly) {
        
-       if (primaryOnly == true && confirmedOnly == true) // TODO: Re-implement that specific case with HQL
```
dkayiwa: Can you throw some more light regarding the removal of this?

mks-d: Until now the case `primaryOnly == true && confirmedOnly == true` was still handled by the legacy implementation. This PR is precisely solving this: this last case is also reimplemented using an HQL query.

mks-d: This would allow to resolve EA-114 since all 4 possible cases for `getUniqueDiagnoses(Boolean, Boolean)` have now been reimplemented using HQL.

dkayiwa: Gat it! :)

PR #119, opened 2016-12-23
Set form field path in observation map

aj-jaswanth:


dkayiwa:
Thanks @aj-jaswanth 👍

PR #120, opened 2017-01-05
EA-116 - Adding an encounter without a prior visit

ssmusoke:
@dkayiwa  I am running into a  strange error with the test in this case. The visit service is not available as its called the stack trace is https://gist.github.com/ssmusoke/5c1bce5b520d5e140a1ca5e3595f8cdf

### DIFF
--- a/api/src/main/java/org/openmrs/module/emrapi/adt/EmrApiVisitAssignmentHandler.java
+++ b/api/src/main/java/org/openmrs/module/emrapi/adt/EmrApiVisitAssignmentHandler.java
@@ -15,6 +15,7 @@
 package org.openmrs.module.emrapi.adt;
 
 import org.apache.commons.lang.StringUtils;
+import org.apache.commons.lang.time.DateUtils;
 import org.openmrs.Encounter;
 import org.openmrs.EncounterType;
 import org.openmrs.Patient;
@@ -115,11 +116,11 @@ public class EmrApiVisitAssignmentHandler extends BaseEncounterVisitHandler impl
             if (visitType != null) {
                 Visit visit = new Visit();
                 visit.setStartDatetime(encounter.getEncounterDatetime());
-                visit.setLocation(encounter.getLocation());
+                visit.setLocation(adtService.getLocationThatSupportsVisits(encounter.getLocation()));
                 visit.setPatient(encounter.getPatient());
                 visit.setVisitType(visitType);
                 //set stop date time to last millisecond of the encounter day for a past visit
-                if (!encounter.getEncounterDatetime().equals(new Date())) {
+                if (!DateUtils.isSameDay(encounter.getEncounterDatetime(), new Date())) {
                     visit.setStopDatetime(OpenmrsUtil.getLastMomentOfDay(encounter.getEncounterDatetime()));
                 }
                 visit.addEncounter(encounter);
--- a/api/src/main/java/org/openmrs/module/emrapi/adt/EmrApiVisitAssignmentHandler.java
+++ b/api/src/main/java/org/openmrs/module/emrapi/adt/EmrApiVisitAssignmentHandler.java
@@ -14,13 +14,23 @@
 
 package org.openmrs.module.emrapi.adt;
 
+import org.apache.commons.lang.StringUtils;
 import org.openmrs.Encounter;
+import org.openmrs.EncounterType;
 import org.openmrs.Patient;
 import org.openmrs.Visit;
+import org.openmrs.VisitType;
+import org.openmrs.api.APIException;
+import org.openmrs.api.AdministrationService;
 import org.openmrs.api.VisitService;
 import org.openmrs.api.context.Context;
 import org.openmrs.api.handler.BaseEncounterVisitHandler;
 import org.openmrs.api.handler.EncounterVisitHandler;
+import org.openmrs.api.handler.ExistingOrNewVisitAssignmentHandler;
+import org.openmrs.module.emrapi.EmrApiConstants;
+import org.openmrs.util.OpenmrsConstants;
+import org.openmrs.util.OpenmrsUtil;
+import org.springframework.beans.factory.annotation.Autowired;
 
 import java.util.Collections;
 import java.util.Date;
@@ -30,13 +40,18 @@ import java.util.Locale;
 /**
  * Ensures that encounters are assigned to visits based on the EMR module's business logic.
  * <p/>
- * For now, we require that a compatible visit exist before you're allowed to create an encounter.
+ * For now, we require that a compatible visit exist before you're allowed to create an encounter. However if the EmrApiConstants#GP_VISIT_ASSIGNMENT_HANDLER_ENCOUNTER_TYPE_TO_VISIT_TYPE_MAP
+ * property that provides a mapping between encounter types and visit types is set then a new visit is created using the visit type from the saved mapping
  */
 public class EmrApiVisitAssignmentHandler extends BaseEncounterVisitHandler implements EncounterVisitHandler {
 
     private VisitService visitService;
 
     private AdtService adtService;
+    
+    private AdministrationService administrationService;
+    
+    private EncounterTypetoVisitTypeMapper encounterTypetoVisitTypeMapper;
 
     /**
      * Since the OpenMRS core doesn't load this bean via Spring, do some hacky setup here.
@@ -48,6 +63,8 @@ public class EmrApiVisitAssignmentHandler extends BaseEncounterVisitHandler impl
             // in production, set the fields this way
             visitService = Context.getVisitService();
             adtService = Context.getService(AdtService.class);
+            administrationService = Context.getAdministrationService();
+            encounterTypetoVisitTypeMapper = Context.getRegisteredComponents(EncounterTypetoVisitTypeMapper.class).get(0);
         } catch (Exception ex) {
             // unit tests will set the fields manually
         }
@@ -91,6 +108,23 @@ public class EmrApiVisitAssignmentHandler extends BaseEncounterVisitHandler impl
                 }
             }
         }
+        // there is no suitable visit so create one if there is a mapping encounter type to the visit type via the Global property
+        if (StringUtils.isNotBlank(administrationService.getGlobalProperty(EmrApiConstants.GP_VISIT_ASSIGNMENT_HANDLER_ENCOUNTER_TYPE_TO_VISIT_TYPE_MAP))) {
+            VisitType visitType = getEncounterTypetoVisitTypeMapper().getVisitTypeForEncounter(encounter);
+            // only process a visit if there is a matching visitType
+            if (visitType != null) {
+                Visit visit = new Visit();
+                visit.setStartDatetime(encounter.getEncounterDatetime());
+                visit.setLocation(encounter.getLocation());
+                visit.setPatient(encounter.getPatient());
+                visit.setVisitType(visitType);
+                //set stop date time to last millisecond of the encounter day for a past visit
+                if (!encounter.getEncounterDatetime().equals(new Date())) {
+                    visit.setStopDatetime(OpenmrsUtil.getLastMomentOfDay(encounter.getEncounterDatetime()));
+                }
+                visit.addEncounter(encounter);
+            }
+        }
 
         // TEMP HACK: allow visit-free encounters while we continue to discuss this
         // throw new IllegalStateException("Cannot create an encounter outside of a visit");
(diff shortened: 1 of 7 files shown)


ssmusoke:
@mogoodrich yes we can I am tracking the comments on JIRA

dkayiwa:
@ssmusoke are you planning to squash these commits into one? :)

api/src/main/java/org/openmrs/module/emrapi/adt/EmrApiVisitAssignmentHandler.java
```diff
@@ -91,6 +103,24 @@ public void beforeCreateEncounter(Encounter encounter) {
...
+            VisitType visitType = encounterTypetoVisitTypeMapper.getVisitTypeForEncounter(encounter);
+            // only process a visit if there is a matching visitType
+            if (visitType != null) {
+                Visit visit = new Visit();
+                visit.setStartDatetime(encounter.getEncounterDatetime());
+                visit.setLocation(encounter.getLocation());
+                visit.setPatient(encounter.getPatient());
+                visit.setVisitType(visitType);
+                //set stop date time to last millisecond of the encounter day.
+                visit.setStopDatetime(OpenmrsUtil.getLastMomentOfDay(encounter.getEncounterDatetime()));
```
djazayeri: I wouldn't set stopDatetime if the visit we're creating is for today.

api/src/main/java/org/openmrs/module/emrapi/adt/EmrApiVisitAssignmentHandler.java
```diff
@@ -91,6 +103,24 @@ public void beforeCreateEncounter(Encounter encounter) {
                 }
             }
         }
+        // there is no suitable visit so create one if there is a mapping encounter type to the visit type via the Global property
+        if (StringUtils.isNotBlank(administrationService.getGlobalProperty(EmrApiConstants.GP_VISIT_ASSIGNMENT_HANDLER_ENCOUNTER_TYPE_TO_VISIT_TYPE_MAP))) {
+            EncounterTypetoVisitTypeMapper encounterTypetoVisitTypeMapper = Context.getRegisteredComponents(EncounterTypetoVisitTypeMapper.class).get(0);
+            encounterTypetoVisitTypeMapper.setMappingString(administrationService.getGlobalProperty(EmrApiConstants.GP_VISIT_ASSIGNMENT_HANDLER_ENCOUNTER_TYPE_TO_VISIT_TYPE_MAP));
```
djazayeri: I haven't finished reading all the code yet but it seems strange to me that we'd fetch a bean, and then set a property on it.

api/src/main/java/org/openmrs/module/emrapi/adt/EmrApiVisitAssignmentHandler.java
```diff
@@ -103,4 +133,6 @@ public void setVisitService(VisitService visitService) {
     public void setAdtService(AdtService adtService) {
         this.adtService = adtService;
     }
+    
+    public void setAdministrationService(AdministrationService administrationService) {this.administrationService = administrationService; }
```
djazayeri: inconsistent formatting

api/src/main/java/org/openmrs/module/emrapi/adt/EncounterTypetoVisitTypeMapper.java
```diff
@@ -0,0 +1,91 @@
...
+ * 2. encounterTypeUuid:visitTypeUuid
+ * 3. A mixture of uuids and id
+ * 4. default:visitTypeId or default:visitTypeUuid which maps all encounter types to the specified visitType
+ *
+ * Any specific mapping will override the default mapping
+ *
+ *
+ */
+@Component
+public class EncounterTypetoVisitTypeMapper {
```
djazayeri: As I understand the logic, every time the VisitAssignmentHandler is called, you fetch the GP value (in case it has changed). If that's the logic you want, then it belongs in this class, i.e. in getVisitTypeForEncounter you'd call a new method like updateMappingsFromGlobalProperty().

api/src/test/java/org/openmrs/module/emrapi/adt/EmrApiVisitAssignmentHandlerTest.java
```diff
@@ -38,21 +45,25 @@
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
-public class EmrApiVisitAssignmentHandlerTest {
+public class EmrApiVisitAssignmentHandlerTest extends BaseModuleContextSensitiveTest {
```
djazayeri: Can't you just mock the administration service as is done for other services, and avoid making this test context-sensitive?

api/src/test/java/org/openmrs/module/emrapi/adt/EmrApiVisitAssignmentHandlerTest.java
```diff
@@ -109,5 +120,82 @@ public void testAssigningASuitableVisitWhenOneExists() throws Exception {
...
+        administrationService.saveGlobalProperty(new GlobalProperty(EmrApiConstants.GP_VISIT_ASSIGNMENT_HANDLER_ENCOUNTER_TYPE_TO_VISIT_TYPE_MAP, ""));
+        
+        handler.beforeCreateEncounter(encounter);
+        
+        // there is a visit on the encounter
+        Assert.assertNull(encounter.getVisit());
+    }
+    
+    @Test
+    public void testCreatingMultipleEncountersWithNewPastVisitsBeingCreatedSpecifiedGlobalProperty() {
```
djazayeri: I didn't look closely but I don't understand the purpose of having a test for "multiple".

Also you should have a test that tests a specific encounter type. (Or maybe that's covered in the tests on the mapper class.

api/src/test/java/org/openmrs/module/emrapi/adt/EncounterTypeToVisitTypeMapperTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.openmrs.EncounterType;
+import org.openmrs.VisitType;
+import org.openmrs.api.context.Context;
+import org.openmrs.test.BaseModuleContextSensitiveTest;
+
+public class EncounterTypeToVisitTypeMapperTest extends BaseModuleContextSensitiveTest {
```
djazayeri: FYI it's better style to do this using mock(EncounterService).

omod/src/main/resources/config.xml
```diff
@@ -123,6 +123,13 @@
             UUID of non coded concept
         </description>
     </globalProperty>
+    <globalProperty>
+        <property>emrapi.EmrApiVisitAssignmentHandler.encounterTypeToNewVisitTypeMap</property>
+        <defaultValue></defaultValue>
+        <description>
+            Specifies the mapping of encounter types to new visit types
```
djazayeri: Ideally you'd give more description about how to set this, though I don't think that will fit here. So maybe document it on the wiki, and then put a link to the wiki here?

ssmusoke:
@djazayeri All updates completed for review

djazayeri:
Skimmed quickly and it seems fine.

One thing I should have mentioned earlier...there's actually GlobalPropertyChangeListener functionality in core that you should be using instead of doing update() every time this is called. (But that's a nice-to-have change, not a required change.)

ssmusoke:
@mogoodrich a ping for your input on the updated PR

mogoodrich:
For what it's worth, I *believe* if you ever get in a state where a patient has two active visits, this causes issues, so we might want to explicitly confirm that this logic can never put us in that state.  (It might be that you can have multiple active visits, as long as they aren't at the same location, I'm not 100% sure)

ssmusoke:
@mogoodrich The new functionality only runs when there is no visit mapped that has ended by the encounter date https://github.com/ssmusoke/openmrs-module-emrapi/blob/fdd1a51ee8db82ac70805cd8a4afed05b1666f38/api/src/main/java/org/openmrs/module/emrapi/adt/EmrApiVisitAssignmentHandler.java#L103-L110 

and when the GP is specified as a mapper. So this is more like a catch for encounters created without visits.

ssmusoke:
@mogoodrich Just a ping to see if you are okay with this PR

mogoodrich:
@ssmusoke  did you see my comments on the EmrApiVisitAssignmentHandler code?  We won't be setting the new GP to start out, so I don't think this will affect us, but just as an fyi, not sure it's doing exactly what you want, but I may be wrong.

ssmusoke:
@mogoodrich I had responded:

1. If you do not set the GP then this functionality is not activated

2. If you set the GP, the new visit is only created when:
a) No prior matching visit exists on the encounter date 

b) There is a mapping for the encounter type to a visit type in the GP

The GP therefore acts like a feature toggle for those who want the functionality

mogoodrich:
aha, sorry @ssmusoke .... I made some notes as part of a review, but I didn't realize I needed to hit "submit review" before they were visible for others... sorry about that!  I just did...  Do you see my two comments now?  Not blockers for us, but worth considering.

mogoodrich:
(And yes, the GP functionality works great for us)

api/src/main/java/org/openmrs/module/emrapi/adt/EmrApiVisitAssignmentHandler.java
```diff
@@ -91,6 +108,23 @@ public void beforeCreateEncounter(Encounter encounter) {
...
             }
         }
+        // there is no suitable visit so create one if there is a mapping encounter type to the visit type via the Global property
+        if (StringUtils.isNotBlank(administrationService.getGlobalProperty(EmrApiConstants.GP_VISIT_ASSIGNMENT_HANDLER_ENCOUNTER_TYPE_TO_VISIT_TYPE_MAP))) {
+            VisitType visitType = getEncounterTypetoVisitTypeMapper().getVisitTypeForEncounter(encounter);
+            // only process a visit if there is a matching visitType
+            if (visitType != null) {
+                Visit visit = new Visit();
+                visit.setStartDatetime(encounter.getEncounterDatetime());
+                visit.setLocation(encounter.getLocation());
```
djazayeri: Here you need to use AdtService.getLocationThatSupportsVisits(encounter.getLocation())

api/src/main/java/org/openmrs/module/emrapi/adt/EmrApiVisitAssignmentHandler.java
```diff
@@ -91,6 +108,23 @@ public void beforeCreateEncounter(Encounter encounter) {
...
+                visit.setPatient(encounter.getPatient());
+                visit.setVisitType(visitType);
+                //set stop date time to last millisecond of the encounter day for a past visit
+                if (!encounter.getEncounterDatetime().equals(new Date())) {
```
djazayeri: Don't assume that the clock hasn't ticked since encounterDatetime, and that these are exactly equal. You'd need to test something like encounterDatetime >= new Date()-1m

ssmusoke: @mogoodrich Changed the comparison to use an apache lang DateUtils method that only compares dates

api/src/main/java/org/openmrs/module/emrapi/adt/EmrApiVisitAssignmentHandler.java
```diff
@@ -91,6 +108,23 @@ public void beforeCreateEncounter(Encounter encounter) {
...
+                    visit.setStopDatetime(OpenmrsUtil.getLastMomentOfDay(encounter.getEncounterDatetime()));
```
ssmusoke: @mogoodrich The fix to the issue is a hibernate interceptor that drops the millisecond bit, which seems to deal with the above named issue.

api/src/main/java/org/openmrs/module/emrapi/adt/EmrApiVisitAssignmentHandler.java
ssmusoke: @djazayeri I have changed to use DateUtils.isSameDay() from Apache lang

api/src/main/java/org/openmrs/module/emrapi/adt/EmrApiVisitAssignmentHandler.java
mogoodrich: @ssmusoke Right, but I believe the interceptor fix is only in OpenMRS 2.0 (at least that's what the fix version for TRUNK-4252).  If you are only using this in 2.0+ you should be fine, but wanted to give you a heads up in case you were using earlier versions.

ssmusoke:
@mogoodrich I saw a comment about the interceptor working for 2.x (which seems to have disappeared) and yes that is my target platform for usage as the 1.11.x is EOL. I can add that additional information to the usage on the WIKI to make it clear

api/src/main/java/org/openmrs/module/emrapi/adt/EmrApiVisitAssignmentHandler.java
ssmusoke: @mogoodrich The target is 2.x not 1.11.x or earlier. I can add that additional information to the usage on the WIKI to make it clear for the target platform

mogoodrich: Sounds good to me, thanks!

mogoodrich:
fyi, I think this is good to be merged (at least from my perspective).

PR #125, opened 2017-03-31
EA-123| Fix wrong endDate on condition status changed

Gaurav-Deshkar:
https://issues.openmrs.org/browse/EA-123


dkayiwa:
@sidtharthanan what do you think of this pull request? :)

sidtharthanan:
@dkayiwa Yes, we raised this pull request.
We need to use onSetDate as the endDate for older condition when it's status change.
We modified this way.

PR #126, opened 2017-04-07
ConditionHistory api throws null pointer exception when a different locale is used

sidtharthanan:
When user is logged in with a locale for which condition concept does not have a name, ConditionHistory api throws NullPointerException.

Issue: [EA-124](https://issues.openmrs.org/browse/EA-124)

### DIFF
--- a/condition-list/src/main/java/org/openmrs/module/emrapi/conditionslist/contract/ConditionMapper.java
+++ b/condition-list/src/main/java/org/openmrs/module/emrapi/conditionslist/contract/ConditionMapper.java
@@ -1,6 +1,7 @@
 package org.openmrs.module.emrapi.conditionslist.contract;
 
 import static org.apache.commons.lang3.StringUtils.isEmpty;
+import static org.openmrs.util.LocaleUtility.getDefaultLocale;
 
 import org.openmrs.ConceptName;
 import org.openmrs.Patient;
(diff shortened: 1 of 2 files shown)


sidtharthanan:
@dkayiwa Could you have a look at it.

condition-list/src/main/java/org/openmrs/module/emrapi/conditionslist/contract/ConditionMapper.java
```diff
@@ -66,8 +67,11 @@ public Condition map(org.openmrs.Condition openmrsCondition) {
 	}
 	
 	private Concept mapConcept(org.openmrs.Concept openmrsConcept) {
-		Concept concept = new Concept(openmrsConcept.getUuid(),
-				openmrsConcept.getFullySpecifiedName(Context.getLocale()).getName());
+		ConceptName fullySpecifiedName = openmrsConcept.getFullySpecifiedName(Context.getLocale());
+		if(fullySpecifiedName == null){
```
dkayiwa: What happens when there is no name in the default locale?

sidtharthanan: It will be null.
Shall we get the first non null FullySpecifiedName, If a FullySpecifiedName not available in the given Locale.
Or do you have any alternate solution?

sidtharthanan: @dkayiwa, Just following up on the above comment. Do you have any inputs?

dkayiwa: Sorry i had forgotten about this. Thanks for the reminder. 😊 
Am merging because this is still better than what we had before.

dkayiwa:
Thanks @sidtharthanan :)

PR #127, opened 2017-06-27
EA-127 | Support for Interpretation and status field in EncounterTran…

shashikanthgadgay:
…saction.Observation

### DIFF
--- a/api-1.10/pom.xml
+++ b/api-1.10/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>org.openmrs.module</groupId>
         <artifactId>emrapi</artifactId>
-        <version>1.22.1</version>
+        <version>1.23.0-SNAPSHOT</version>
     </parent>
 
     <artifactId>emrapi-api-1.10</artifactId>
--- a/api-1.11/pom.xml
+++ b/api-1.11/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>org.openmrs.module</groupId>
         <artifactId>emrapi</artifactId>
-        <version>1.22.1</version>
+        <version>1.23.0-SNAPSHOT</version>
     </parent>
 
     <artifactId>emrapi-api-1.11</artifactId>
--- a/api-1.12/pom.xml
+++ b/api-1.12/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>org.openmrs.module</groupId>
         <artifactId>emrapi</artifactId>
-        <version>1.22.1</version>
+        <version>1.23.0-SNAPSHOT</version>
     </parent>
 
     <artifactId>emrapi-api-1.12</artifactId>
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -4,7 +4,7 @@
 	<parent>
 		<groupId>org.openmrs.module</groupId>
 		<artifactId>emrapi</artifactId>
-		<version>1.22.1</version>
+		<version>1.23.0-SNAPSHOT</version>
 	</parent>
 
 	<artifactId>emrapi-api</artifactId>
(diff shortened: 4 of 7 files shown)


api-2.1/src/main/java/org/openmrs/module/emrapi/encounter/mapper/DrugMapper2_1.java
```diff
@@ -0,0 +1,38 @@
...
+import org.springframework.stereotype.Component;
+
+@Component(value = "drugMapper")
+@OpenmrsProfile(openmrsPlatformVersion = "2.1.*")
+public class DrugMapper2_1 implements DrugMapper {
+
+    public EncounterTransaction.Drug map(Drug drug){
+        EncounterTransaction.Drug encounterTransactionDrug = new EncounterTransaction.Drug();
+        encounterTransactionDrug.setName(drug.getDisplayName());
+        if (drug.getDosageForm() != null) {
```
dkayiwa: How does this class differ from DrugMapper1_12?

api-2.1/src/main/java/org/openmrs/module/emrapi/encounter/mapper/OpenMRSOrderMapper.java
```diff
@@ -0,0 +1,111 @@
...
+import org.openmrs.module.emrapi.CareSettingType;
+import org.openmrs.module.emrapi.encounter.domain.EncounterTransaction;
+
+import java.util.Iterator;
+
+/**
+ * OpenMRSOrderMapper.
+ * Maps EncounterTransaction Order to OpenMRS Orders.
+ * <p/>
+ * Version 1.0
```
dkayiwa: What does Version 1.0 stand for?

dkayiwa:
On a side note, are all these changes for just the two new fields that were introduced?
Yes know reflection is generally discouraged, but in cases where it is a few lines in comparison to these massive changes, it looks like an attractive alternative, to me. 😊

shashikanthgadgay:
Yes, All these changes are for two new fields. As they are part of OpenMRS 2.1, I had to create new module. 

I have duplicated few files so that they are supported in the 2.1 version of OpenMRS.
https://github.com/openmrs/openmrs-module-emrapi/pull/127/commits/7036792ab1b717df2265c2eaf01c3fc711fd19e5#diff-0e8ec544c2dc4fee40711c8b5f5c8f29R33

shashikanthgadgay:
I can refactor the code by using reflection. But someone has to add new module 2.1 at some point of time. 
What do you suggest to do now?

dkayiwa:
My personal opinion is that if reflection requires only like one percent of the total changes in this pull request, i would go for it. It is not a must that we shall need the 2.1 module. Just like we do not have the 2.0 module. Did you notice that? 😊 

The problem with some of these sub modules that are addressing only minor changes, is that they involve lots of duplication and hence more code to maintain.

shashikanthgadgay:
Ya I agree that it involves a lot of duplication. I will refactor the code by using reflection and update the pull request

shashikanthgadgay:
@dkayiwa 
I tried to refactor code by using reflection, But as the two fields are of Enum type. The code has become a bit complex. 
I added the below code for one of the fields. 

`
 Class<Enum> interpretationClass = (Class<Enum>) Class.forName("org.openmrs.Obs$Interpretation");
 Method setInterpretation = obs.getClass().getMethod("setInterpretation", new Class[] { interpretationClass});
 Enum interpretation = Enum.valueOf(interpretationClass, observationData.getInterpretation());
 setInterpretation.invoke(obs, interpretation);
`

Does this look good to you?

dkayiwa:
If that is all you have to do in comparison to the original massive changes, then i would go for it. I would put it in a private method whose name says what is going on. And i would add some javadoc to the method.

shashikanthgadgay:
@dkayiwa I have refactored the code. Was wondering how do I write unit tests for it?

dkayiwa:
@shashikanthgadgay that is a very good question. 😊 
Now this is where your initial approach of the 2.1 sub module gets easier.
But i have a feeling that you may have copied more files than needed. For i noticed a few which were identical.

PR #130, opened 2017-12-29
[EA-125] Add a "diagnosis" REST resource

BartlomiejRasztabiga:
Issue: https://issues.openmrs.org/browse/EA-125

### DIFF
--- a/omod/src/main/java/org/openmrs/module/emrapi/web/controller/DiagnosisController.java
+++ b/omod/src/main/java/org/openmrs/module/emrapi/web/controller/DiagnosisController.java
@@ -1,0 +1,37 @@
+package org.openmrs.module.emrapi.web.controller;
+
+import java.util.Date;
+import java.util.List;
+
+import org.openmrs.Patient;
+import org.openmrs.module.emrapi.diagnosis.Diagnosis;
+import org.openmrs.module.emrapi.diagnosis.DiagnosisService;
+import org.openmrs.module.webservices.rest.web.v1_0.controller.BaseRestController;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.ResponseBody;
+
+@Controller
+@RequestMapping(value = "/rest/emrapi")
+public class DiagnosisController extends BaseRestController {
+
+	@Autowired
+	private DiagnosisService diagnosisService;
+
+	@RequestMapping(method = RequestMethod.GET, value = "/patientdiagnoses")
+	@ResponseBody
+	public List<Diagnosis> getDiagnosesList(@RequestParam("patient") Patient patient,
+											@RequestParam(value = "fromDate", required = false) Date fromDate) {
+		return diagnosisService.getDiagnoses(patient, fromDate);
+	}
+
+	@RequestMapping(method = RequestMethod.GET, value = "/patientuniquediagnoses")
+	@ResponseBody
+	public List<Diagnosis> getUniqueDiagnosesList(@RequestParam("patient") Patient patient,
+			                                      @RequestParam(value = "fromDate", required = false) Date fromDate) {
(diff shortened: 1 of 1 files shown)


omod/src/main/java/org/openmrs/module/emrapi/web/controller/DiagnosisController.java
```diff
@@ -0,0 +1,37 @@
...
+@Controller
+@RequestMapping(value = "/rest/emrapi")
+public class DiagnosisController extends BaseRestController {
+
+	@Autowired
+	private DiagnosisService diagnosisService;
+
+	@RequestMapping(method = RequestMethod.GET, value = "/diagnosislist")
+	@ResponseBody
+	public List<Diagnosis> getDiagnosesList(@RequestParam("patient") Patient patient,
```
dkayiwa: Should the fromDate be required?

omod/src/main/java/org/openmrs/module/emrapi/web/controller/DiagnosisController.java
```diff
@@ -0,0 +1,37 @@
...
+											@RequestParam("fromDate") Date fromDate) {
+		return diagnosisService.getDiagnoses(patient, fromDate);
+	}
+
+	@RequestMapping(method = RequestMethod.GET, value = "/uniquediagnosislist")
+	@ResponseBody
+	public List<Diagnosis> getUniqueDiagnosesList(@RequestParam("patient") Patient patient,
+			                                      @RequestParam("fromDate") Date fromDate) {
```
dkayiwa: Should the fromDate be required?

omod/src/main/java/org/openmrs/module/emrapi/web/controller/DiagnosisController.java
BartlomiejRasztabiga: It is required by getDiagnoses() method

omod/src/main/java/org/openmrs/module/emrapi/web/controller/DiagnosisController.java
BartlomiejRasztabiga: It is required by getUniqueDiagnoses() method

omod/src/main/java/org/openmrs/module/emrapi/web/controller/DiagnosisController.java
```diff
@@ -0,0 +1,37 @@
...
+import org.springframework.web.bind.annotation.ResponseBody;
+
+@Controller
+@RequestMapping(value = "/rest/emrapi")
+public class DiagnosisController extends BaseRestController {
+
+	@Autowired
+	private DiagnosisService diagnosisService;
+
+	@RequestMapping(method = RequestMethod.GET, value = "/diagnosislist")
```
dkayiwa: Also from the talk thread, Darius seems to suggest "patientdiagnoses" for the url

dkayiwa:
And by the way, have you tried to run the module with these changes and made this rest call?

omod/src/main/java/org/openmrs/module/emrapi/web/controller/DiagnosisController.java
dkayiwa: The api allows it to be null.

omod/src/main/java/org/openmrs/module/emrapi/web/controller/DiagnosisController.java
dkayiwa: The api allows it to be null.

BartlomiejRasztabiga:
Do I have to download all these modules to run EMR API? 
image

ivange94:
You can find the dependencies for emrapi here https://github.com/openmrs/openmrs-module-emrapi/blob/master/omod/src/main/resources/config.xml#L20-L28

And yes, you need to install all it's dependencies to run it. Note that, it's dependencies may also have dependencies of their own.

BartlomiejRasztabiga:
And how can I know which version of these modules should I install? Always newest?

ivange94:
The link above tells which module versions you need but just that it uses already defined variables. To know the value of these variables you can look at the properties section of the parent pom of a module. For example for this module you can find them here https://github.com/openmrs/openmrs-module-emrapi/blob/master/pom.xml#L40-L55

BartlomiejRasztabiga:
Thanks a lot :)

PR #134, opened 2018-01-05
EA-131| EmrConceptSearchController should allow search by concept reference term code

ShruthiPitta:
https://issues.openmrs.org/browse/EA-131

### DIFF
--- a/omod/src/main/java/org/openmrs/module/emrapi/web/controller/EmrConceptSearchController.java
+++ b/omod/src/main/java/org/openmrs/module/emrapi/web/controller/EmrConceptSearchController.java
@@ -15,8 +15,10 @@
 package org.openmrs.module.emrapi.web.controller;
 
 import org.openmrs.Concept;
-import org.openmrs.ConceptName;
+import org.openmrs.ConceptMap;
+import org.openmrs.ConceptReferenceTerm;
 import org.openmrs.ConceptSearchResult;
+import org.openmrs.ConceptSource;
 import org.openmrs.module.emrapi.EmrApiProperties;
 import org.openmrs.module.emrapi.concept.EmrConceptService;
 import org.openmrs.module.webservices.rest.SimpleObject;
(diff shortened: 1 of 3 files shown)


omod/src/main/java/org/openmrs/module/emrapi/web/controller/EmrConceptSearchController.java
```diff
@@ -43,28 +45,42 @@
...
+        List<ConceptSource> conceptSources = emrApiProperties.getConceptSourcesForDiagnosisSearch();
         Locale locale = Locale.ENGLISH;
-        List<ConceptSearchResult> conceptSearchResults = emrService.conceptSearch(query, locale, null, diagnosisSets, null, limit);
-        List<ConceptName> matchingConceptNames = new ArrayList<ConceptName>();
-        for (ConceptSearchResult searchResult : conceptSearchResults) {
-            matchingConceptNames.add(searchResult.getConceptName());
-        }
-        return createListResponse(matchingConceptNames);
+        List<ConceptSearchResult> conceptSearchResults = emrService.conceptSearch(query, locale, null, diagnosisSets, conceptSources, limit);
+        return createListResponse(conceptSearchResults, conceptSources.get(0));
```
dkayiwa: Is see that you are passing conceptSources.get(0)
What happens when conceptSources has more than one source?

ShruthiPitta: @dkayiwa It is returning the ICD10 concept source converting it to list https://github.com/openmrs/openmrs-module-emrapi/blob/5296fb6c3ea2e382a14195cb00b0212f42f4762c/api/src/main/java/org/openmrs/module/emrapi/EmrApiProperties.java#L189. However we may have to refactor it to return a concept source which is configured as a global property.

dkayiwa: Do you know why it is returning it to a list when it is actually one?

ShruthiPitta: No. I do not know. However there is another issue: https://issues.openmrs.org/browse/EA-130. As part of this we could refactor it?

dkayiwa: Can we make this code in such a way that it does not assume one item in the list? That way, when it gets fixed to return more than one item, this code does not need to be changed.

ShruthiPitta: I think the fix will still return only one concept source. If you want me to change the method signature to return one concept source I can do that.

dkayiwa: What puzzles me is why return a list when expecting only one item? Could there be something we are missing that the original author had in mind?

ShruthiPitta: ok. May be @djazayeri can help us here. @djazayeri  Do you know why EmrApiProperties.getConceptSourcesForDiagnosisSearch returns a list though we have are looking only for ICD10 source? Or may be you just moved the code for some refactoring?

mks-d: @ShruthiPitta and @dkayiwa, I asked this [here](https://talk.openmrs.org/t/ui-for-diagnosis-search/14834/9?u=mksd) on Talk.

dkayiwa: Thanks @mks-d 😊

PR #135, opened 2018-02-27
EA-133, ignore voided encounters

cioan:


cioan:
Thanks @SilverDragon5000 for reviewing this change. My intent was to disregard voided encounters when searching for encounters in the last given number of hours.

PR #159, opened 2018-11-29
Update openmrsVersion to 2.2.0-alpha

samuelmale:


omod/pom.xml
```diff
@@ -426,7 +426,7 @@
         <profile>
             <id>MRS2.0</id>
             <properties>
-                <openMRSVersion>2.0.1-SNAPSHOT</openMRSVersion>
+                <openMRSVersion>2.2.0-alpha</openMRSVersion>
```
dkayiwa: No. This should be the latest release in the 2.1.x series

PR #160, opened 2018-12-03
Updated the openmrsVersion dependency to 2.1.3

samuelmale:


omod/pom.xml
```diff
@@ -426,7 +426,7 @@
         <profile>
             <id>MRS2.0</id>
             <properties>
-                <openMRSVersion>2.0.1-SNAPSHOT</openMRSVersion>
```
dkayiwa: 2.1.3 is not in the 2.0.x series

samuelmale: Ooohh sorry. I wasn't notified of the edit you made here https://github.com/openmrs/openmrs-module-emrapi/pull/159#discussion_r237479054

PR #162, opened 2019-03-02
EA-141: Fixed randomly failing test

HelioStrike:
The test AdtServiceComponentTest.test_getVisitsAndHasVisitDuring() failed randomly because of the line: 
```
assertFalse(service.hasVisitDuring(patient, outpatientDepartment, stopDate, now));
```
where stopDate is some date from 2012. There was no test above this line which created a visit between the 2 dates, hence the test wasn't definitely failing. As the test patient was being fetched from the database, there's no telling if the patient doesn't have a visit between the specified date range(hasVisitDuring() internally fetches inactive visits too!), which caused the test to fail 'randomly'. I wasn't sure what I should be replacing the line with, so I simply removed it. Doing that fixes the issue.


dkayiwa:
The original author intentionally put that assertion. So just removing it is not the fix.

HelioStrike:
Got it. Would you suggest adding an 'includeInactive' parameter to getVisits() (would require changing the ~10 function call) or copying the function code into the test?

HelioStrike:
@dkayiwa  I added 2 new methods to AdtService, namely getActiveVisits() and hasActiveVisitDuring().  
The line which was failing:
```
assertFalse(service.hasVisitDuring(patient, outpatientDepartment, stopDate, now));
```
now becomes:
```
assertFalse(service.hasActiveVisitDuring(patient, outpatientDepartment, stopDate, now));
```
I ran `mvn clean install` several times to make sure that the test passes, and it seems to be working. I guess that fixes the issue?

HelioStrike:
@dkayiwa I added 2 new methods to AdtService, namely getActiveVisits() and hasActiveVisitsDuring().
The line which was causing the test to fail:
```
assertFalse(service.hasVisitDuring(patient, outpatientDepartment, stopDate, now));
```
now becomes:
```
assertFalse(service.hasActiveVisitDuring(patient, outpatientDepartment, stopDate, now));
```
I also changed the line:
```
Date now = new Date(System.currentTimeMillis() - 60*1000);
```
to:
```
Date now = new Date(System.currentTimeMillis() - 60*1000);
```
, going back by a minute. I think going back by a second would work too, but during creation visits timings are precise up to a minute, so I went with that option. I tested the new code by running `mvn clean install` around 10 times, but the test passes.

dkayiwa:
Do you have to add the new methods to fix the randomly failing test?

HelioStrike:
@dkayiwa I would assume so. I believe the test is trying to check that the number of active visits until the current point in time is 0. Since hasVisitsDuring() returns all visits, I think the new methods would serve the purpose better.

api/src/main/java/org/openmrs/module/emrapi/adt/AdtService.java
```diff
@@ -300,6 +300,29 @@ VisitDomainWrapper createRetrospectiveVisit(Patient patient, Location location,
      */
     boolean hasVisitDuring(Patient patient, Location location, Date startDatetime, Date stopDatetime);
 
+    /**
+     * Gets all visits for the patient at the visit location associated with the specified location
+     * during the specified datetime range
+     *
+     * @param patient
+     * @param location
```
dkayiwa: Can you document these params?

api/src/main/java/org/openmrs/module/emrapi/adt/AdtService.java
```diff
@@ -300,6 +300,29 @@ VisitDomainWrapper createRetrospectiveVisit(Patient patient, Location location,
      */
     boolean hasVisitDuring(Patient patient, Location location, Date startDatetime, Date stopDatetime);
 
+    /**
+     * Gets all visits for the patient at the visit location associated with the specified location
```
dkayiwa: The description does not seem to communicate what is reflected by the method name.

HelioStrike:
@dkayiwa I've updated the comments. Please have another look.

api/src/main/java/org/openmrs/module/emrapi/adt/AdtService.java
```diff
@@ -300,6 +300,30 @@ VisitDomainWrapper createRetrospectiveVisit(Patient patient, Location location,
...
+     * @return
+     */
+    List<VisitDomainWrapper> getActiveVisits(Patient patient, Location location, Date startDatetime, Date stopDatetime);
+
+    /**
+     * Returns true/false whether or not the patient has any active visits at the visit location associated with
+     * the specified location during the specified datetime range
+     *
+     * @param patient
+     * @param location
```
dkayiwa: Can you document these params?

dkayiwa:
@HelioStrike are you still working on this?

HelioStrike:
@dkayiwa I think its ready and good to go. I'm not sure how to document the params though. Here's the hasVisitDuring() function (present before the changes):
```
    /**
     * Returns true/false whether or not the patient has any visits at the visit location associated with
     * the specified location during the specified datetime range
     *
     * @param patient
     * @param location
     * @param startDatetime
     * @param stopDatetime
     * @return
     */
    boolean hasVisitDuring(Patient patient, Location location, Date startDatetime, Date stopDatetime);
```
And here's the new function, hasActiveVisitDuring():
```
    /**
     * Returns true/false whether or not the patient has any active visits at the visit location associated with
     * the specified location during the specified datetime range
     *
     * @param patient
     * @param location
     * @param startDatetime
     * @param stopDatetime
     * @return
     */
    boolean hasActiveVisitDuring(Patient patient, Location location, Date startDatetime, Date stopDatetime);
```
All I did was change the description a little to match what the new function was doing. Is there any thing I'm supposed to do?

dkayiwa:
> Should I go ahead and put a comment beside each param?

Yes go ahead.

PR #163, opened 2019-03-17
EA-91: Deprecated MetadataUtil class in EMR API, 

mozzy11:
Ticket https://issues.openmrs.org/browse/EA-91
"EA-91: Remove MetadataUtil class from EMR API, removed the tests , and also removed the dependence on Metadasharing Module"

### DIFF
--- a/api/src/main/java/org/openmrs/module/emrapi/utils/MetadataUtil.java
+++ b/api/src/main/java/org/openmrs/module/emrapi/utils/MetadataUtil.java
@@ -47,7 +47,16 @@ import java.util.TreeMap;
 import java.util.TreeSet;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
-
+/*@deprecated .This class has been moved to the openmrs Metadata Sharing Module
+ * Ensure you have at least version 1.6.0 of Metadata Sharing Module
+ * Change this to call class org.openmrs.module.metadatasharing.packages.MetadataUtil.java
+ * Change your metadata files to use aliased name <MetadataPackagesConfig> 
+ *      rather than the full name <org.openmrs.module.emrapi.metadata.MetadataPackagesConfig>
+ * Change your metadata files to use aliased name <MetadataPackageConfig>
+ *      rather than the full name <org.openmrs.module.emrapi.metadata.MetadataPackageConfig>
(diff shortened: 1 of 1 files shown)


mozzy11:
hi @djazayeri  , checkout this PR. i instead just deprecated the Util class in Emrapi and added an alias to the xtream Serializer and also added a test

api/src/main/java/org/openmrs/module/emrapi/utils/MetadataUtil.java
```diff
@@ -48,6 +49,10 @@
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+/**
+ * @deprecated . Should delegate to org/openmrs/module/metadatasharing/util/MetadataUtil.java
```
djazayeri: Make this comment better. Need to say that:

* this method has moved
* you need to do two things:
    1. ensure you have at least version xyz of other module xyz
    1. change this to call method xyz of other module xyz
    1. change your metadata files to use xyz format
* more info at (link to JIRA ticket)

api/src/main/java/org/openmrs/module/emrapi/utils/MetadataUtil.java
```diff
@@ -130,8 +135,10 @@ public static MetadataPackagesConfig getMetadataPackagesForModule(ClassLoader lo
 		try {
 			InputStream stream = loader.getResourceAsStream(packageFilePath);
 			String xml = IOUtils.toString(stream);
-			MetadataPackagesConfig config = Context.getSerializationService().getDefaultSerializer()
-					.deserialize(xml, MetadataPackagesConfig.class);
+			SimpleXStreamSerializer serializer = (SimpleXStreamSerializer) Context.getSerializationService().getDefaultSerializer();
```
djazayeri: Don't make this change here. We only want to support two things:

1. User doesn't make any changes, things keep working as they were
2. User makes all the documented changes at once.

The only changes in this module should be adding deprecation notices, and instructions.

djazayeri:
Please also add deprecation notices on MetadataPackagesConfig and MetadataPackageConfig

mozzy11:
@djazayeri , i have worked on your comments. only added the deprecation and the notices

gracepotma:
@dkayiwa can you review this?

PR #167, opened 2019-07-23
EA-151 : diagnosis name search by concept sets to take precedence over b…

Ruhanga:
…oth concept classes and sources restrictions

Ticket : https://issues.openmrs.org/browse/EA-151

cc : @mks-d, @mogoodrich

### DIFF
--- a/api/src/main/java/org/openmrs/module/emrapi/concept/EmrConceptDAO.java
+++ b/api/src/main/java/org/openmrs/module/emrapi/concept/EmrConceptDAO.java
@@ -32,6 +32,18 @@ public interface EmrConceptDAO {
 
     List<Concept> getConceptsMappedTo(Collection<ConceptMapType> mapTypes, ConceptReferenceTerm term);
 
+    /**
+     * Searches for concepts based on either inSets or classes, sources or combination of classes and sources 
+     * in the specified locale. inSets overrides sources and classes if provided.
+     * Can search concepts based on code for the query restricted by the sources provided.
+     *
+     * @param query name or term of concept to search for
+     * @param locale locale to search in
+     * @param classes concept classes to search against
+     * @param inSets concept sets to search in
+     * @param sources concept source to searcch against
+     * @param limit the maximum results to fetch
+     */
     List<ConceptSearchResult> conceptSearch(String query, Locale locale, Collection<ConceptClass> classes, Collection<Concept> inSets, Collection<ConceptSource> sources, Integer limit);
 
 }
--- a/api/src/main/java/org/openmrs/module/emrapi/concept/EmrConceptDAO.java
+++ b/api/src/main/java/org/openmrs/module/emrapi/concept/EmrConceptDAO.java
@@ -34,8 +34,8 @@ public interface EmrConceptDAO {
 
     /**
      * Searches for concepts based on either inSets or classes, sources or combination of classes and sources 
-     * in the specified locale. inSets overrides sources and classes if provided.
-     * Can search concepts based on code for the query restricted by the sources provided.
+     * in the specified locale. If inSets is specified, it overrides sources and classes.
+     * Can search concepts based on code as the query restricted by the specified sources.
      *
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/openmrs/module/emrapi/concept/HibernateEmrConceptDAO.java
```diff
@@ -91,25 +91,27 @@ public void setSessionFactory(DbSessionFactory sessionFactory) {
...
             if (inSets != null) {
                 DetachedCriteria allowedSetMembers = DetachedCriteria.forClass(ConceptSet.class);
                 allowedSetMembers.add(Restrictions.in("conceptSet", inSets));
                 allowedSetMembers.setProjection(Projections.property("concept"));
                 criteria.add(Subqueries.propertyIn("concept", allowedSetMembers));
             }
 
-            for (String word : uniqueWords) {
-                criteria.add(Restrictions.ilike("name", word, MatchMode.ANYWHERE));
+            if (!CollectionUtils.isEmpty(classes) && CollectionUtils.isEmpty(inSets)) {
```
mogoodrich: I think it would be worth adding a comment here (or in the javadocs for the method) noting that it only restricts by class if no set specified.

api/src/main/java/org/openmrs/module/emrapi/concept/HibernateEmrConceptDAO.java
```diff
@@ -91,25 +91,27 @@ public void setSessionFactory(DbSessionFactory sessionFactory) {
...
+                conceptCriteria.add(Restrictions.in("conceptClass", classes));
             }
 
-            if (!CollectionUtils.isEmpty(sources)) {
-            	Criteria mappingCriteria = conceptCriteria.createCriteria("conceptMappings");
+            if (!CollectionUtils.isEmpty(sources) && CollectionUtils.isEmpty(inSets)) {
```
mogoodrich: And a similar comment here to what I mentioned above.

api/src/main/java/org/openmrs/module/emrapi/concept/HibernateEmrConceptDAO.java
Ruhanga: Thank you @mogoodrich, I have a java doc to the method.

api/src/main/java/org/openmrs/module/emrapi/concept/EmrConceptDAO.java
```diff
@@ -32,6 +32,18 @@
 
     List<Concept> getConceptsMappedTo(Collection<ConceptMapType> mapTypes, ConceptReferenceTerm term);
 
+    /**
+     * Searches for concepts based on either inSets or classes, sources or combination of classes and sources 
+     * in the specified locale. If inSets is specified, it overrides sources and classes.
+     * Can search concepts based on code as the query restricted by the specified sources.
```
mks-d: >Searches for concepts by name either 1) within the specified concept sources and/or concept classes, or 2) within the specified concept sets.
The name search within classes and/or sources is bypassed when sets are provided for the search. In that case the name search only operates within those sets, and the classes and/or sources are just ignored.
Also searches for concepts by mapping code if the concept sources are specified (regardless whether classes and/or sets are specified.)

api/src/main/java/org/openmrs/module/emrapi/concept/EmrConceptDAO.java
```diff
@@ -32,6 +32,18 @@
...
+     *
+     * @param query name or term of concept to search for
+     * @param locale locale to search in
+     * @param classes concept classes to search against
+     * @param inSets concept sets to search in
+     * @param sources concept source to searcch against
```
mks-d: Typo: `searcch`

Ruhanga: Cool, thanks @mks-d.

PR #169, opened 2019-08-22
RA-1488: Error while saving condition

haripriya999:
# RA-1488 : Error while saving condition

## Description of what I changed
Fix error while saving condition

## Issue I worked on

see https://issues.openmrs.org/browse/RA-1488

## Checklist: I completed these to help reviewers :)

- [x] My pull request only contains **ONE single commit**
(the number above, next to the 'Commits' tab is 1).

  No? -> [read here](https://wiki.openmrs.org/display/docs/Pull+Request+Tips) on how to squash multiple commits into one

- [x] My IDE is configured to follow the [**code style**](https://wiki.openmrs.org/display/docs/Java+Conventions) of this project.

  No? Unsure? -> [configure your IDE](https://wiki.openmrs.org/display/docs/How-To+Setup+And+Use+Your+IDE), format the code and add the changes with `git add . && git commit --amend`

- [x] I have **added tests** to cover my changes. (If you refactored
  existing code that was well tested you do not have to add tests)

  No? -> write tests and add them to this commit `git add . && git commit --amend`

- [x] I ran `mvn clean package` right before creating this pull request and
  added all formatting changes to my commit.

  No? -> execute above command

- [x] All new and existing **tests passed**.

  No? -> figure out why and add the fix to your commit. It is your responsibility to make sure your code works.

- [x] My pull request is **based on the latest changes** of the master branch.

  No? Unsure? -> execute command `git pull --rebase upstream master`

### DIFF
--- a/web-2.2/src/main/java/org/openmrs/module/emrapi/web/controller/ConditionController.java
+++ b/web-2.2/src/main/java/org/openmrs/module/emrapi/web/controller/ConditionController.java
@@ -8,6 +8,7 @@ import java.util.Collection;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Locale;
 
 import org.openmrs.CodedOrFreeText;
 import org.openmrs.Concept;
@@ -140,6 +141,9 @@ public class ConditionController extends BaseRestController {
 		ConceptName fullySpecifiedName = coreConcept.getFullySpecifiedName(Context.getLocale());
 		if (fullySpecifiedName == null) {
 			fullySpecifiedName = coreConcept.getFullySpecifiedName(getDefaultLocale());
+		}
+		if (fullySpecifiedName == null) {
+			fullySpecifiedName = coreConcept.getFullySpecifiedName(new Locale("en"));
 		}
 		org.openmrs.module.emrapi.conditionslist.contract.Concept concept =
 				new org.openmrs.module.emrapi.conditionslist.contract.Concept(coreConcept.getUuid(),


web-2.2/src/main/java/org/openmrs/module/emrapi/web/controller/ConditionController.java
```diff
@@ -141,6 +142,9 @@ private ConditionClinicalStatus convertConditionListStatus(Condition.Status stat
 		if (fullySpecifiedName == null) {
 			fullySpecifiedName = coreConcept.getFullySpecifiedName(getDefaultLocale());
 		}
+		if (fullySpecifiedName == null) {
+			fullySpecifiedName = coreConcept.getFullySpecifiedName(new Locale("en"));
```
dkayiwa: Why do we have this and the one before?

haripriya999: @dkayiwa  In the server logs it was showing a null pointer exception as no locale was being set on line 143.  So if the application is not able to fetch the default locale, then I'm assigning the locale to "en"

dkayiwa: Did you find out why the default locale was not returned in the fresh set up?

haripriya999: @dkayiwa I realized that it is not the locale that is being set to null, but the getFullySpecifiedName() on line 142 here: https://github.com/openmrs/openmrs-module-emrapi/blob/master/web-2.2/src/main/java/org/openmrs/module/emrapi/web/controller/ConditionController.java#L142 that is returning null. On inspecting openmrs-core, I see that here on line 699: https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/Concept.java#L699 the value of locale is 'en_GB' but this if statement over here: https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/Concept.java#L704 is returning false, so it gets out of the outer if statement. The value of nameLocale on line 713 here: https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/Concept.java#L713 is being set to "en" because of which the if condition over here: https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/Concept.java#L714 is failing as locale.getLanguage() gives "en_GB" and nameLocale.getLanguage() gives "en" and the function is returning null value.

dkayiwa: Can you explain all the above classes and lines you have pointed out, in relation to when it worked well on your local version? I want to see the difference between the two.

haripriya999: @dkayiwa  I not really sure why its happening this way, but when I tested it on local version(Reff App 2.9.0 standalone) this if statement here: https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/Concept.java#L704 returned concept name, hence the fullySpecifiedName here: https://github.com/openmrs/openmrs-module-emrapi/blob/master/web-2.2/src/main/java/org/openmrs/module/emrapi/web/controller/ConditionController.java#L142 was not null, but for some reason its showing different behavior on Reff App 2.10.0-SNAPSHOT

dkayiwa: Can you debug both and find the cause?

haripriya999: @dkayiwa  I'm getting this error while deploying openmrs war file on server https://pastebin.com/xY6XXHJ6 . How can this be fixed? I changed the version of war.openmrs in openmrs-server properties file, yet I get the error

dkayiwa: How do your changes look like?

haripriya999: @dkayiwa  I added System.out.println's for testing purposes  . Also on comparing 2.1.x and 2.2.x I see that we have some difference over here. https://github.com/openmrs/openmrs-core/blob/2.2.x/api/src/main/java/org/openmrs/ConceptName.java#L495 (2.2.x) and https://github.com/openmrs/openmrs-core/blob/2.1.x/api/src/main/java/org/openmrs/ConceptName.java#L489 (2.1.x) . I'm not really sure if this is the very cause, so I am trying to make some conclusions using System.out.println's

haripriya999: @dkayiwa when locale is set to en_GB getNames(locale).isEmpty() is being set to true and this line: https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/Concept.java#L700 is returning false and we get the error. It's happening because, here on this line: https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/Concept.java#L736 getLocale() is returning "en" instead of "en_GB" and the value of variable locale is "en_GB", and hence its returning an empty list causing the error. I'm not able to understand why getLocale() is returning "en" as this function is supposed to return the current locale which is "en_GB" . I'm trying my best to figure out things here. Can we merge this PR as a temporary patch for fixing this bug till then?

haripriya999: @dkayiwa  Through print statements, I am able to conclude that on Openmrs 2.1.x even though the default locale is set to "en_GB" somehow the locale is being changed to "en" before it goes to this function : https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/Concept.java#L699 , So the getFullySpecifiedName() function, is taking the locale as "en" though the default locale is "en_GB"  and hence is working. On deliberately testing the function with "en_GB" , I'm able to produce the same error on 2.1.x as well. On OpenMRS 2.2.x platform, the getFullySpecifiedName() function takes "en",only when the locale is set to "en" , so we are able to generate this error on 2.2.x. with "en_GB" as locale. I feel this PR can be merged to fix it, as there seems to be no issue with the core.

PR #176, opened 2019-12-05
RA-1641: inactive conditions should be displayed

mozzy11:
Ticket : https://issues.openmrs.org/browse/RA-1641

added a handler to return all Conditions


mozzy11:
cc @mseaton

mozzy11:
true @mogoodrich , let work on that

mozzy11:
Hi @mogoodrich  , i have updated the PR as per your suggestion

PR #177, opened 2020-01-23
RA-950: Add LastVisit and EditVisit require an EndDate

reagan-meant:
https://issues.openmrs.org/browse/RA-950
I did the following
1.Enabled adding a past visit with a start date but no end date
2.But disabled adding a past visit with a start date but no end date if this would overlap with another existing visit
3.Enabled editing a closed visit and remove its end date
4.But disabled editing a closed visit and remove its end date if this would overlap with another existing visit
Other changes are in core-apps https://github.com/openmrs/openmrs-module-coreapps/pull/275

### DIFF
--- a/api/src/main/java/org/openmrs/module/emrapi/adt/AdtServiceImpl.java
+++ b/api/src/main/java/org/openmrs/module/emrapi/adt/AdtServiceImpl.java
@@ -784,16 +784,16 @@ public class AdtServiceImpl extends BaseOpenmrsService implements AdtService {
     @Transactional
     public VisitDomainWrapper createRetrospectiveVisit(Patient patient, Location location, Date startDatetime, Date stopDatetime)
         throws ExistingVisitDuringTimePeriodException {
-
+        
         if (startDatetime.after(new Date())) {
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/openmrs/module/emrapi/adt/AdtServiceImpl.java
```diff
@@ -784,7 +784,10 @@ public VisitDomainWrapper wrap(Visit visit) {
     @Transactional
     public VisitDomainWrapper createRetrospectiveVisit(Patient patient, Location location, Date startDatetime, Date stopDatetime)
         throws ExistingVisitDuringTimePeriodException {
-
+        Date actualStopDate = stopDatetime;
+        if (stopDatetime == null){
+            stopDatetime = new Date();
+        }
```
mogoodrich: Is this just to avoid a null pointer exception on line 792? If so, I would just add a null check there, that is:

  if (if stopDatetime != null && stopDatetime.after(new Date())) {

reagan-meant: Thanx @mogoodrich I have adjusted

PR #180, opened 2020-03-06
MAP-19: Setting a UUID to the extra PIT metadata set.

mks-d:


mogoodrich:
Actually, if it's easy enough (I think it is?) can you add code to fetch the metadata set by it's code and then update the uuid if necessary, so that implementations won't have to write their own migrations? @mks-d

mks-d:
@mogoodrich where do you see this to happen?

mogoodrich:
I would think just within the createExtraPatientIdTypesSetIfMissing method, ie if extraPatientIdTypesMapping != null then update the uuid, though I guess this would a bit dangerous.  Though the flip side is that all existing implementatoins would have a uuid that doesn't match the constant.

mks-d:
@mogoodrich FYI @Ruhanga will do the follow-up commit as per your suggestion.

PR #181, opened 2020-03-12
EA-154| EmrApiVisitAssignmentHandler; Check for Visit Id in associated visit to make sure it is a saved visit

mddubey:
Jira issue: https://issues.openmrs.org/browse/EA-154
OpenMRS Talk: https://talk.openmrs.org/t/emrapivisitassignmenthandler-should-check-for-null-visitid/27359

### DIFF
--- a/api/src/test/java/org/openmrs/module/emrapi/adt/EmrApiVisitAssignmentHandlerTest.java
+++ b/api/src/test/java/org/openmrs/module/emrapi/adt/EmrApiVisitAssignmentHandlerTest.java
@@ -297,4 +297,37 @@ public class EmrApiVisitAssignmentHandlerTest {
         Assert.assertThat(encounter.getEncounterDatetime(), is(suitable.getStopDatetime()));
     }
 
+    @Test
+    public void testIgnoreTheAssociatedVisitIfItIsNotYetSaved() {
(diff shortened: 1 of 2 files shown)


api/src/test/java/org/openmrs/module/emrapi/adt/EmrApiVisitAssignmentHandlerTest.java
```diff
@@ -297,4 +297,37 @@ public void testAdjustingEncounterTimeToEndOfVisitWhenAssignedToVisit() throws E
...
+
+        Visit associatedVisit = new Visit();
+        associatedVisit.setStartDatetime(DateUtils.addDays(new Date(), -1));
+        associatedVisit.setLocation(location);
+        encounter.setVisit(associatedVisit);
+
+        handler.beforeCreateEncounter(encounter);
+        encounter.setEncounterDatetime(encounterDatetime);
+
+        Assert.assertThat(encounter.getVisit(), is(suitable));
```
mogoodrich: This seems wrong... if a visit has been associated with an encounter we *want* to keep that visit associated with the encounter, regardless of whether or not it has been persisted, correct?

mddubey:
Closing the PR as it is not applicable.

PR #185, opened 2020-06-01
EA-158: Allow loading disposition configs from file system

mogoodrich:
https://issues.openmrs.org/browse/EA-158


mogoodrich:
Not sure why the other commits ended up here, but the diff looks okay.

PR #191, opened 2020-08-11
Update to event-api-2.8.0

ibacher:
This should hopefully fix failing builds on Bamboo


mogoodrich:
@ibacher I just tried removing the api-1.x and api-2.x dependencies and the tests passed on my local machine that is running the problematic version of Java!

ibacher:
@mogoodrich The 2.10 version of the RefApp runs 2.7.0. And woohoo! That's great!

ibacher:
I guess I'll have to make a proper PR instead of the simple edit I did.

mogoodrich:
Okay, cool... I will hope that there was no particular reason the PIH EMR was running Event 2.5 rather than we forgot to upgrade.

PR #193, opened 2020-10-27
EA-163: Add condition to bypass saveVisit() when existing visit.

binduak:
#### Ticket:
https://issues.openmrs.org/browse/EA-163

#### What I have done:
Added condition to ignore `saveVisit()` for the existing visit.


api/src/main/java/org/openmrs/module/emrapi/encounter/EmrEncounterServiceImpl.java
```diff
@@ -326,6 +331,8 @@ private Visit findOrCreateVisit(EncounterTransaction encounterTransaction, Patie
         visit.setStartDatetime(getCurrentDateIfNull(encounterTransaction.getEncounterDateTime()));
         visit.setEncounters(new HashSet<Encounter>());
         visit.setUuid(UUID.randomUUID().toString());
-        return visit;
```
mks-d, line 329: Instead of all these changes, why not just change
```java
return visit;
```
by
```java
return visitService.saveVisit(visit);
```
?

binduak: sure @mks-d the suggested changes looks more sense to me. I will also test the flow with the new changes and get back to you

binduak: @mks-d requesting you to review the latest changes. Thanks !

mks-d (commented):
@binduak I'd expect to see a test that validates the new behaviour, as in a test that checks that `saveVisit(Visit)` is never invoked if the visit isn't new.

binduak:
> @binduak I'd expect to see a test that validates the new behaviour, as in a test that checks that `saveVisit(Visit)` is never invoked if the visit isn't new.

@mks-d Added test case. Please review. Thanks !

mks-d (approved):

PR #196, opened 2021-03-13
EA-165 | Searching in user locale context. Also supporting locale in parameter

angshu:


api/src/main/resources/messages.properties
```diff
@@ -4,6 +4,7 @@ emrapi.retrospectiveVisit.startDateCannotBeInFuture=Visit start date cannot be i
 emrapi.retrospectiveVisit.stopDateCannotBeInFuture=Visit stop date cannot be in the future
 emrapi.retrospectiveVisit.endDateBeforeStartDateMessage=Visit end date can't be before start date
 emrapi.retrospectiveVisit.patientAlreadyHasVisit=Patient already has visit during that time period
+emrapi.conceptSearch.invalidLocale=Locale is either not valid or not in allowed locales list
```
dkayiwa, line 7: Wouldn't it be easier for the consumer if these are two separate messages?

dkayiwa: Including the invalid value in the error message looks more friendly too. 😊

angshu: Updated. had to fetch the message for parameterisation, APIException does not have means of parameterised message for that version of core.

dkayiwa: @angshu looks good to me! 👍

dkayiwa: It would be great to include the ticket id in the commit message.

angshu: Added the number to the commit description. If we squash and merge, then it can go as single description.

PR #205, opened 2021-08-17
EA-170: Visit Domain Wrapper: add method to return whether visit has …

mogoodrich:
…been discharged


brandones (approved):
Seems reasonable to me!

PR #208, opened 2023-02-22
EA-173: "o3" privileges should be considered application-level privil…

mogoodrich:
…eges

### DIFF
--- a/api/src/main/java/org/openmrs/module/emrapi/EmrApiConstants.java
+++ b/api/src/main/java/org/openmrs/module/emrapi/EmrApiConstants.java
@@ -86,6 +86,8 @@ public class EmrApiConstants {
 
     public static final String PRIVILEGE_PREFIX_TASK = "Task: ";
 
+    public static final String PRIVILEGE_PREFIX_O3 = "o3";
+
     public static final String PRIVILEGE_DELETE_ENCOUNTER = "Task: emr.patient.encounter.delete";
 
     public static final String PRIVILEGE_EDIT_ENCOUNTER = "Task: emr.patient.encounter.edit";
--- a/api/src/main/java/org/openmrs/module/emrapi/account/AccountServiceImpl.java
+++ b/api/src/main/java/org/openmrs/module/emrapi/account/AccountServiceImpl.java
@@ -194,7 +194,8 @@ public class AccountServiceImpl extends BaseOpenmrsService implements AccountSer
 
     private boolean isApplicationPrivilege(Privilege privilege) {
         return privilege.getPrivilege().startsWith(EmrApiConstants.PRIVILEGE_PREFIX_APP)
-                || privilege.getPrivilege().startsWith(EmrApiConstants.PRIVILEGE_PREFIX_TASK);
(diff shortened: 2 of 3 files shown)


mogoodrich:
This is pretty straightforward, but we shouldn't merge until we confirm the naming convention we want to use for "o3" privileges.

api/src/main/java/org/openmrs/module/emrapi/EmrApiConstants.java
```diff
@@ -86,6 +86,8 @@ public class EmrApiConstants {
 
     public static final String PRIVILEGE_PREFIX_TASK = "Task: ";
 
+    public static final String PRIVILEGE_PREFIX_O3 = "o3";
+
```
mseaton, line 90: I don't really see why we don't keep the same "App:" and "Task:" permissions.  Why should the same functionality require different permissions based on choice of front-end technology?  If I copied a page from a GSP over to an O3 microfrontend, I would expect to be able to use the same privileges as before.

mseaton (commented):
See my comment.  I think we need more discussion on this potentially @mogoodrich .  @ibacher has this already been decided?

api/src/main/java/org/openmrs/module/emrapi/EmrApiConstants.java
mogoodrich: It's definitely not already been decided and the naming convention is the main thing I wanted feedback on. I'd be happy to keep the same privilege privilege naming convention, I just thought there'd be  more likelihood to get buy-in on something that starts with "o3" rather than "App: " and "Task: "...  @ibacher  @mks-d  who are the right people to bring together to make a consensus about this?

Though, thinking about it now, I guess I'm tending to lean even more towards just starting with "App: " and "Task: " task within dispensing... if another convention is picked up for privileges in o3, we could migrate the Dispensing privileges to whatever is chosen, but at least "App: " and "Task: " is consistent with OpenMRS 2 in the meantime.

mks-d: @mogoodrich I think that would be @ibacher :-)

mogoodrich: @mks-d I added you because of what said here: https://talk.openmrs.org/t/o3-privileges-plan/36038

ibacher: Personally, I think keeping the `App: ` convention makes sense. Since O3 isn’t a straightforward port of O2 anyways, we’ll likely end up with somewhat different app names anyways, but as much as possible, it makes sense to let implementations leverage their existing privileges where they make sense…

mogoodrich: Cool, I'm thinking I'll move towards this approach (using "App:" and "Task:") based on @ibacher  and @mseaton thoughts...

mogoodrich:
Closing this out as "won't fix" and have switched the Dispensing app to use the OpenMRS 2.0 "Task:" style privileges.

PR #211, opened 2023-03-24
RA-1678 - Update on timezone fix that moves date interpretation to th…

mseaton:
…e server endpoint

This PR changes the model object that is used to handle the JSON representation of a condition such that the onSet and endDate properties are Strings and not Dates.  This ensures that there is no String->Date conversion happening in the Spring/Jackson internals, as I have as of yet been unsuccessful in trying to configure that to work as I'd expect.  Instead, this moves the Date conversion into a separate step that happens when this contract Condition is converted into an actual Condition.  This allows us full control over the Date conversion process.

### DIFF
--- a/condition-list/src/main/java/org/openmrs/module/emrapi/conditionslist/DateConverter.java
+++ b/condition-list/src/main/java/org/openmrs/module/emrapi/conditionslist/DateConverter.java
@@ -1,0 +1,64 @@
+package org.openmrs.module.emrapi.conditionslist;
+
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.joda.time.DateTime;
+import org.joda.time.format.DateTimeFormat;
+
+import java.text.SimpleDateFormat;
+import java.util.Arrays;
+import java.util.Date;
+import java.util.List;
+
+/**
+ * This converter mimics the behavior in the rest webservices ConversionUtil method for dates
+ * This ensures that the system timezone is used to interpret dates unless UTC is explicitly indicated
+ * This is implemented as a
+ */
+public class DateConverter {
+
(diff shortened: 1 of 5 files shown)


ibacher (approved):
LGTM

condition-list/src/main/java/org/openmrs/module/emrapi/conditionslist/DateConverter.java
```diff
@@ -0,0 +1,64 @@
...
+
+import java.text.SimpleDateFormat;
+import java.util.Arrays;
+import java.util.Date;
+import java.util.List;
+
+/**
+ * This converter mimics the behavior in the rest webservices ConversionUtil method for dates
+ * This ensures that the system timezone is used to interpret dates unless UTC is explicitly indicated
+ * This is implemented as a
```
mogoodrich, line 17: Missing comment?

condition-list/src/main/java/org/openmrs/module/emrapi/conditionslist/DateConverter.java
```diff
@@ -0,0 +1,64 @@
...
+import org.joda.time.format.DateTimeFormat;
+
+import java.text.SimpleDateFormat;
+import java.util.Arrays;
+import java.util.Date;
+import java.util.List;
+
+/**
+ * This converter mimics the behavior in the rest webservices ConversionUtil method for dates
+ * This ensures that the system timezone is used to interpret dates unless UTC is explicitly indicated
```
mogoodrich, line 16: isn't this really "unless a timezone is explicity indicate"  (ie, doesn't necessarily need to be UTC)

mogoodrich (commented):
@mseaton  see my comments on the comments, but otherwise looks good....

@ibacher @mks-d as far as we know, Core Apps is the only place that uses these endpoints, correct (ie not Bahmni?)

PR #219, opened 2024-01-04
BAH-3448 | Add. Accession number field in Order class

riyaTw:
Jira Card link - https://bahmni.atlassian.net/browse/BAH-3448?atlOrigin=eyJpIjoiYjZjZTNmOWRkMWJjNDY1ODhlZGI5YjViYWE4MzA5MzAiLCJwIjoiaiJ9

### DIFF
--- a/api/src/main/java/org/openmrs/module/emrapi/encounter/domain/EncounterTransaction.java
+++ b/api/src/main/java/org/openmrs/module/emrapi/encounter/domain/EncounterTransaction.java
@@ -656,10 +656,16 @@ public class EncounterTransaction {
             return autoExpireDate;
         }
 
+        /**
+         * @since 1.35.0
+         */
         public String getAccessionNumber() {
             return accessionNumber;
         }
 
+        /**
+         * @since 1.35.0
+         */
         public void setAccessionNumber(String accessionNumber) {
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/openmrs/module/emrapi/encounter/domain/EncounterTransaction.java
```diff
@@ -655,6 +656,14 @@ public Date getAutoExpireDate() {
             return autoExpireDate;
         }
 
+        public String getAccessionNumber() {
+            return accessionNumber;
+        }
+
```
dkayiwa, line 662: Can you add @since annotations for these new public methods?

riyaTw: @dkayiwa which version should we mention ? 1.34.0 or 1.35.0 ?

dkayiwa: The one that is going to be released with these changes.

riyaTw: Done

PR #222, opened 2024-03-07
EA-181: Relax EMR API dependency on Metadata Sharing

mogoodrich:


mseaton (approved):
Assuming you have tested this and find that it still works as expected with MDS and also starts up and works without errors without MDS, looks good to me.  Not sure why we didn't do this 10 years ago if so.

dkayiwa (approved):

mogoodrich:
Thanks @dkayiwa  and @mseaton !

mogoodrich:
FWIW @mseaton  I didn't test that the deprecated MDS functionality in the module still worked, but I don't see why it wouldn't...

PR #224, opened 2024-05-01
RA-2036 - Provide a mechanism for modules to extend the merge visits …

mseaton:
…functionality via actions

https://openmrs.atlassian.net/browse/RA-2036

### DIFF
--- a/api/src/main/java/org/openmrs/module/emrapi/adt/AdtService.java
+++ b/api/src/main/java/org/openmrs/module/emrapi/adt/AdtService.java
@@ -228,6 +228,7 @@ public interface AdtService extends OpenmrsService {
      * in service impls greatly slow down startup, instead we have other modules inject the beans explicitly
      * via this method)
      *
+     * @since 1.36.0
      * @param visitMergeAction
      */
     void addVisitMergeAction(VisitMergeAction visitMergeAction);
@@ -236,6 +237,7 @@ public interface AdtService extends OpenmrsService {
      * Allows another module to remove a visit merge action to the list of actions to be performed
      * when a visit is merged; (see addVisitMergeAction above)
      *
+     * @since 1.36.0
      * @param visitMergeAction
      */
     void removeVisitMergeAction(VisitMergeAction visitMergeAction);
--- a/api/src/main/java/org/openmrs/module/emrapi/adt/AdtService.java
+++ b/api/src/main/java/org/openmrs/module/emrapi/adt/AdtService.java
@@ -24,6 +24,7 @@ import org.openmrs.Visit;
 import org.openmrs.api.OpenmrsService;
 import org.openmrs.module.emrapi.adt.exception.ExistingVisitDuringTimePeriodException;
 import org.openmrs.module.emrapi.merge.PatientMergeAction;
+import org.openmrs.module.emrapi.merge.VisitMergeAction;
 import org.openmrs.module.emrapi.visit.VisitDomainWrapper;
 
 import java.util.Date;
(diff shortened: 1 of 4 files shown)


mogoodrich (changes requested):
LGTM, the only issue is that I can't find ticket RA-2306 in OpenMRS Jira, so either there's an issue with the search, or there was a typo in the ticket name.  If so, can you correct when merging?  Thanks!

dkayiwa:
https://openmrs.atlassian.net/browse/RA-2036

api/src/main/java/org/openmrs/module/emrapi/adt/AdtService.java
```diff
@@ -221,6 +222,24 @@ Encounter checkInPatient(Patient patient, Location where, Provider checkInClerk,
...
 
+    /**
+     * Allows another module to add a visit merge action to the list of actions to be performed
+     * when a visit is merged; (ideally, this would just happen via @Autowired, but because @Autowired
+     * in service impls greatly slow down startup, instead we have other modules inject the beans explicitly
+     * via this method)
+     *
+     * @param visitMergeAction
+     */
+    void addVisitMergeAction(VisitMergeAction visitMergeAction);
```
dkayiwa, line 233: What do you think of adding the `@since` annotation to these new methods?

api/src/main/java/org/openmrs/module/emrapi/adt/AdtServiceImpl.java
```diff
@@ -747,6 +770,22 @@ public void removePatientMergeAction(PatientMergeAction patientMergeAction) {
...
+    public void addVisitMergeAction(VisitMergeAction visitMergeAction) {
+        if (this.visitMergeActions == null) {
+            this.visitMergeActions = new ArrayList<VisitMergeAction>();
+        }
+        this.visitMergeActions.add(visitMergeAction);
+    }
+
+    @Override
+    public void removeVisitMergeAction(VisitMergeAction visitMergeAction) {
+        if (this.visitMergeActions == null) {
```
dkayiwa, line 783: Do we intentionally ignore cases where one tries to remove a visit merge action when we do not have any?

api/src/main/java/org/openmrs/module/emrapi/adt/AdtServiceImpl.java
```diff
@@ -121,6 +124,11 @@ public List<PatientMergeAction> getPatientMergeActions() {
         return patientMergeActions;
     }
 
+    // for testing
+    public List<VisitMergeAction> getVisitMergeActions() {
```
dkayiwa, line 128: How about getting rid of this method and use reflection to access the property in the test?

api/src/main/java/org/openmrs/module/emrapi/adt/AdtService.java
mseaton: Added @dkayiwa

api/src/main/java/org/openmrs/module/emrapi/adt/AdtServiceImpl.java
mseaton: @dkayiwa I just followed exactly the same pattern as is already in place for the PatientMergeActions.  I figured it was best to stay consistent.

api/src/main/java/org/openmrs/module/emrapi/merge/VisitMergeAction.java
```diff
@@ -0,0 +1,30 @@
...
+     * This method will be called before calling the visit service to save / void either of the given visits,
+     * but in the same transaction. Any thrown exception will cancel the merge
+     *
+     * @param preferred the Visit to keep
+     * @param notPreferred the Visit to merge into the preferred visit
+     */
+    void beforeSavingVisits(Visit preferred, Visit notPreferred);
+
+    /**
+     * This method will be after before calling the visit service to save / void either of the given visits,
```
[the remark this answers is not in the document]
mseaton: Good catch, fixed.

api/src/main/java/org/openmrs/module/emrapi/adt/AdtServiceImpl.java
mseaton: Again, this is just copying the existing pattern in place.  I've changed the method scope to be narrower, so it is no longer a public method.  but I see no harm in this method being here...

dkayiwa (approved):

PR #226, opened 2024-05-23
EA-185 - Remove legacy condition list functionality for emrapi 2.x

mseaton:


mogoodrich (approved):
Seems correct, as I believe this is all in core now?

dkayiwa (approved):

ibacher:
I actually seem to recall that there was some additional logic I left in `saveCondition()` in core to continue supporting this endpoint, so it may be that there's actually more code we can dispense with eventually.

PR #229, opened 2024-05-29
EA-184 - Move reporting module to an aware_of dependency

mseaton:


api-reporting/pom.xml
```diff
@@ -0,0 +1,111 @@
...
+    <parent>
+        <groupId>org.openmrs.module</groupId>
+        <artifactId>emrapi</artifactId>
+        <version>2.0.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>emrapi-api-reporting</artifactId>
+    <packaging>jar</packaging>
+    <name>EMR API Reporting API</name>
+
```
mseaton, line 14: Here, I create a new maven sub-module, which will result in a jar that can be conditionally loaded.  All of the reporting-related dependencies are able to be moved in here.

api-reporting/src/main/java/org/openmrs/module/emrapi/adt/reporting/evaluator/AwaitingAdmissionVisitQueryEvaluator.java
```diff
@@ -0,0 +1,49 @@
...
+import org.openmrs.module.reporting.query.visit.evaluator.VisitQueryEvaluator;
+import org.springframework.beans.factory.annotation.Autowired;
+
+import java.util.Collection;
+import java.util.List;
+
+@Handler(supports = AwaitingAdmissionVisitQuery.class)
+@OpenmrsProfile(modules = { "reporting:*" })
+public class AwaitingAdmissionVisitQueryEvaluator implements VisitQueryEvaluator {
+
```
mseaton, line 23: The logic for retrieving visits awaiting admission was moved into the AdtService so that it can be used without relying on the reporting module.  Then, the reporting module evaluator below was updated to call the implementation in the AdtService.  None of the unit tests were changed, and all continue to pass.

api/pom.xml
```diff
@@ -12,56 +12,13 @@
 	<name>EMR API Module API</name>
 	<description>API project for EMRAPI</description>
 
-    <dependencies>
```
mseaton, line 15: This is residual cleanup from prior refactor, where these dependencies can live in the top level pom and used by all submodules.

api/src/main/java/org/openmrs/module/emrapi/adt/AdtService.java
```diff
@@ -336,4 +337,12 @@ VisitDomainWrapper createRetrospectiveVisit(Patient patient, Location location,
...
     List<Location> getInpatientLocations();
 
+    /**
+     * returns patients awaiting admission
+     * @param location - if supplied, only checks waiting admissions at the visit location associated with this location
+     * @param patientIds - if non-null, only returns matches for patients with the given ids
+     * @param visitIds - if non-null, only returns matches for visits with the given ids
+     * @return List<Visit></Visit> of the matching visits
+     */
+    List<Visit> getVisitsAwaitingAdmission(Location location, Collection<Integer> patientIds, Collection<Integer> visitIds);
```
mseaton, line 347: This is the new AdtService method that now contains the logic that was previously implemented in the reporting module

api/src/main/java/org/openmrs/module/emrapi/db/EmrApiDAO.java
```diff
@@ -0,0 +1,11 @@
+package org.openmrs.module.emrapi.db;
+
+import java.util.List;
+import java.util.Map;
+
```
mseaton, line 5: I was able to get rid of the EmrVisitDAO, which only really existed to execute hql queries, and replaced it with this new EmrApiDAO that provides more general purpose HQL execution methods.  The logic that was in the EmrVisitDAO methods has been moved to the service layer.

api/src/main/java/org/openmrs/module/emrapi/diagnosis/ObsGroupDiagnosisService.java
```diff
@@ -196,23 +198,55 @@ public List<Diagnosis> getUniqueDiagnoses(Patient patient, Date fromDate) {
 		return diagnoses;
 	}
 
+    public List<Obs> getDiagnoses(Visit visit, DiagnosisMetadata diagnosisMetadata) {
```
mseaton, line 201: Previously these were delegating to specific implementations in the EmrVisitDAO.  I moved the logic in here and got rid of the EmrVisitDAO.

api/src/main/java/org/openmrs/module/emrapi/patient/PatientDomainWrapper.java
```diff
@@ -75,9 +74,6 @@ public class PatientDomainWrapper implements DomainWrapper {
 	@Qualifier("emrDiagnosisService")
 	@Autowired
 	protected DiagnosisService diagnosisService;
-
```
mseaton, line 78: Unused

api/src/main/java/org/openmrs/module/emrapi/visit/VisitDomainWrapper.java
```diff
@@ -82,13 +75,13 @@ public enum SortOrder {
     @Autowired
     protected DispositionService dispositionService;
 
-    @Autowired
-    protected VisitQueryService visitQueryService;
-
```
mseaton, line 87: I was able to remove usage of reporting module constructs here, and use the new Adt service method instead, so this no longer relies on the reporting module code.

api/src/main/resources/hql/patients_diagnoses.hql
```diff
@@ -6,5 +6,5 @@ where
     p.voided = 'false'
     and p.patientId in(select distinct personId from Obs o
     where
```
mseaton, line 8: There were some errors / inconsistencies in some of these hql queries, where on one side of a condition it was looking for an id, and the other side an object.  I changed them to consistently refer to objects rather than deconstruct to ids.

api/src/main/resources/hql/visits_awaiting_admission.hql
```diff
@@ -0,0 +1,36 @@
+select
```
mseaton, line 1: This is actually the main thing that was needed to move off of the reporting module dependency.  I moved what was previously being done in a reporting module evaluator using the HqlQueryBuilder, and built a similar query here using HQL directly and the new EmrApiDAO methods that evaluate it.

api/src/test/java/org/openmrs/module/emrapi/diagnosis/ObsGroupDiagnosisServiceComponentTest.java
```diff
@@ -56,11 +56,20 @@ public void setUp() throws Exception {
...
 	}
+	
+	private Date parseYmd(String ymd) {
+		try {
+			return new SimpleDateFormat("yyyy-MM-dd").parse(ymd);
+		}
+		catch (Exception e) {
+			throw new RuntimeException(e);
+		}
+	}
```
mseaton, line 67: This is to eliminate the use of a reporting module utility method

omod/pom.xml
```diff
@@ -13,18 +13,20 @@
 	<description>OMOD project for EMRAPI</description>
 
 	<dependencies>
+
 		<dependency>
```
mseaton, line 17: This is partially clean up of a prior commit that removed the condition-list submodule, and partially part of this new commit that adds the reporting api submodule jar to the omod.

api/src/main/java/org/openmrs/module/emrapi/visit/VisitDomainWrapper.java
```diff
@@ -56,9 +51,7 @@
 import java.util.List;
 import java.util.Map;
 
-import static java.util.Collections.EMPTY_LIST;
-import static java.util.Collections.reverseOrder;
-import static java.util.Collections.sort;
+import static java.util.Collections.*;
```
dkayiwa, line 54: https://openmrs.atlassian.net/wiki/spaces/docs/pages/25477044/Java+Conventions#JavaConventions-Donotusewildcardimports

api/src/test/java/org/openmrs/module/emrapi/diagnosis/OldDiagnosisBuilder.java
```diff
@@ -31,4 +33,13 @@ public ObsBuilder buildDiagnosis(Patient patient, String dateYmd, Diagnosis.Orde
 		}
 		return builder;
 	}
+
+	private Date parseYmd(String ymd) {
```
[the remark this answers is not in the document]
ibacher: Seems like the goal here is to not depend on the `DateUtil` class from the reporting module outside of the new api-reporting jar. That said, `common-lang` has a `DateUtils` class where this could be swapped for `DateUtils.parseDate(dateYmd, "yyyy-MM-dd")`.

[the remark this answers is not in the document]
ibacher: It might be easier to swap to using common-langs `DateUtils#parseDate()`, e.g., `DateUtils.parseDate(dateYmd, "yyyy-MM-dd")` should work. 😁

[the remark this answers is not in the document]
mseaton: Sure, will do.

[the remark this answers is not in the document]
mseaton: Actually that's a different version of DateUtils that we have available and it doesn't really seem to buy anything, so I'm going to leave it as is.

PR #230, opened 2024-05-29
EA-186 - Loosen providermanagement to an aware_of dependency

mseaton:


api/src/main/java/org/openmrs/module/emrapi/account/provider/CoreProviderService.java
```diff
@@ -0,0 +1,49 @@
...
+import org.openmrs.annotation.OpenmrsProfile;
+import org.openmrs.api.ProviderService;
+import org.springframework.beans.factory.annotation.Autowired;
+
+import java.util.Collection;
+
+/**
+ * Provides an abstraction over Provider services
+ */
+@OpenmrsProfile(modules = {"!providermanagement"})
```
mseaton, line 15: Provides an implementation of provider-related methods that are used when providermanagement is not installed.  This would be where we'd implement logic to support setting roles on the core `Provider` object as `Concept role`, but that can be added in a separate PR if desired since that was not pre-existing functionality here.

api/src/main/java/org/openmrs/module/emrapi/account/provider/ProviderManagementProviderService.java
```diff
@@ -0,0 +1,97 @@
+package org.openmrs.module.emrapi.account.provider;
```
mseaton, line 1: Provides an implementation of provider-related methods that are used when providermanagement _is_ installed. This allows us to keep any references to providermanagement hidden behind this conditionally loaded bean.  The `setProviderRole` method is the way it is as the method signature changed from `ProviderRole` to `Object`, and this allows supporting a wide range of possible values including bind parameters in controller requests.

api/src/main/java/org/openmrs/module/emrapi/account/AccountDomainWrapper.java
```diff
@@ -125,21 +122,23 @@ public Provider getProvider() {
         return provider;
     }
 
-    public void setProviderRole(ProviderRole providerRole) {
-
```
mseaton, line 129: This is a necessary backwards-incompatibility on the get/set providerRole methods by broadening to Object as ProviderRole is not available when the providermanagement module is not installed.

api/src/test/java/org/openmrs/module/emrapi/EmrApiContextSensitiveTest.java
```diff
@@ -0,0 +1,24 @@
...
+
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.openmrs.module.Module;
+import org.openmrs.module.ModuleFactory;
+import org.openmrs.test.BaseModuleContextSensitiveTest;
+
+import java.io.File;
+
+public abstract class EmrApiContextSensitiveTest extends BaseModuleContextSensitiveTest {
```
mseaton, line 11: For now, I just made all of the context-sensitive unit tests in the module extend this, which adds the providermanagement module in so that the module-specific version of the facade is activated, as the existing unit tests expect that to be there.  If there is a straightforward to support both variations in the same unit test suite, please let me know.  I was not able to get it working.

api/src/main/java/org/openmrs/module/emrapi/account/provider/CoreProviderService.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    @Override
+    public Provider saveProvider(Provider provider) {
+        return providerService.saveProvider(provider);
+    }
+
+    @Override
+    public Object getProviderRole(Provider provider) {
+        return null;
```
dkayiwa, line 42: Shouldn't this also throw the same exception as setProviderRole?

api/src/main/java/org/openmrs/module/emrapi/visit/VisitDomainWrapper.java
```diff
@@ -51,9 +51,7 @@
 import java.util.List;
 import java.util.Map;
 
-import static java.util.Collections.EMPTY_LIST;
-import static java.util.Collections.reverseOrder;
-import static java.util.Collections.sort;
+import static java.util.Collections.*;
```
dkayiwa, line 54: https://openmrs.atlassian.net/wiki/spaces/docs/pages/25477044/Java+Conventions#JavaConventions-Donotusewildcardimports

api/src/test/java/org/openmrs/module/emrapi/EmrApiContextSensitiveTest.java
dkayiwa: Just like you created `BaseReportingTest` in the previous pull request, shouldn't you rename this to something like `BaseProvidermanagementTest`?

api/src/main/java/org/openmrs/module/emrapi/visit/VisitDomainWrapper.java
mogoodrich: Strange, I feel like my version of IntelliJ started doing wildcard imports again as well after I turned it off years ago.

api/src/main/java/org/openmrs/module/emrapi/account/provider/CoreProviderService.java
mseaton: The way I was seeing it, provider role is always null if the providermanagement module is not running.  So why throw an exception?  The setter is a different situation - the user is trying to set something and the module is unable to accommodate.

api/src/test/java/org/openmrs/module/emrapi/EmrApiContextSensitiveTest.java
mseaton: Because I make this the superclass of every context-sensitive test in the module, it felt more appropriate to name it this way.

api/src/main/java/org/openmrs/module/emrapi/visit/VisitDomainWrapper.java
mseaton: Fixed.

PR #231, opened 2024-06-17
O3-3193: REST API for EMR API ADT functionality

mogoodrich:


mogoodrich (commented):
So after rabbit-holing a bit I pulled back and did the bare-bones simplest stuff to get the two major REST queries we will need for the Ward app as a proof-of-concept, see attached.

Certainly not ready for merging, and we may go another way, but wanted to get this out in draft form since I'm out tomorrw... @chibongho  @mseaton

omod/src/main/java/org/openmrs/module/emrapi/web/controller/InpatientVisitsController.java
```diff
@@ -0,0 +1,74 @@
...
+        // TODO handle null response if possible
+        // TODO this getInpatentVisits method is almost certainly not performant enough for production use and will likely need to be refactored into a HQL query
+        List<VisitDomainWrapper> visits = adtService.getInpatientVisits(adtService.getLocationThatSupportsVisits(currentLocation), currentLocation);
+
+        // TODO type this?
+        List<SimpleObject> response = new ArrayList<SimpleObject>();
+
+        for (VisitDomainWrapper visit : visits) {
+            SimpleObject inpatientVisit = new SimpleObject();
+            inpatientVisit.put("visit", ConversionUtil.convertToRepresentation(visit.getVisit(), Representation.DEFAULT));
```
mseaton, line 44: It would be better if we could support different representations, right?  Not sure if you can just take in "rep" as an additional request parameter and then apply it?  Or possibly the most straightforward way to do this would be to type the response object as a class named "InpatientVisit", and then implement this as a `@Resource` with `supportedClass = InpatientVisit.class`rather than just a a custom `@Controller`?

omod/src/main/java/org/openmrs/module/emrapi/web/controller/InpatientVisitsController.java
```diff
@@ -0,0 +1,74 @@
...
+            inpatientVisit.put("patient", ConversionUtil.convertToRepresentation(visit.getVisit().getPatient(), Representation.DEFAULT));
+            inpatientVisit.put("currentLocation", ConversionUtil.convertToRepresentation(currentLocation, Representation.DEFAULT));
+            inpatientVisit.put("timeSinceAdmissionInMinutes",  Minutes.minutesBetween(new DateTime(visit.getAdmissionEncounter().getEncounterDatetime()), new DateTime()).getMinutes());
+            inpatientVisit.put("timeOnWardInMinutes",  Minutes.minutesBetween(new DateTime(visit.getLatestAdtEncounter().getEncounterDatetime()), new DateTime()).getMinutes());  // TODO: assumption, an ADT ecounter always results in change of ward?
```
mseaton, line 48: So this wouldn't cover the (probably rare) case of a bed swap within the same ward, assuming a TRANSFER encounter is used for this.  Though if a different encounter type were used, then this would work.  I tried to describe this situation in the ticket, but it wasn't very clear, and it's something we can discuss

omod/src/main/java/org/openmrs/module/emrapi/web/controller/InpatientVisitsController.java
```diff
@@ -0,0 +1,74 @@
...
+    @ResponseBody
+    public List<SimpleObject> getVisitsAwaitingAdmission(@RequestParam("admissionLocation") Location admissionLocation) {
+
+        // TODO note that this service method *only* returns admission requests, while we will need to expand this to include transfer requests (which will be slightly non-trivial)
+        // TODO note also that this service method does *not* actually limit by admission location; we will need to expand the underlying service method/hql query to do this
+        List<Visit> visits = adtService.getVisitsAwaitingAdmission(admissionLocation, null, null);
+        List<SimpleObject> response = new ArrayList<SimpleObject>();
+        for (Visit visit : visits) {
+            SimpleObject inpatientVisit = new SimpleObject();
+            inpatientVisit.put("visit", ConversionUtil.convertToRepresentation(visit, Representation.DEFAULT));
```
mseaton, line 65: Same comment as above about supporting different representations.  Seems like a reasonable first pass, in the sense that this will provide some quasi-legitimate data to start, though as you indicate there are several issues.

omod/src/main/java/org/openmrs/module/emrapi/web/controller/InpatientVisitsController.java
mogoodrich: Yep, that would be a limitation.  I just dropped in those "timeSince" and "timeWard" values during my last few minutes of work on Monday to has some examples of parameter, (and there some satisfaction in seeing them work without much extra work!).   I think we are going to need to rewrite the underlying service method using a query, and we *may* want to tackle calculating these values as part of that, but if we don't, it should should be very straightforward to move these two time calculations out of the controller and into the VDW and handle the edge case you describe there by iterating over encounters and comparing locations.  In fact, easy enough that it probably makes sense for me to move them now.

omod/src/main/java/org/openmrs/module/emrapi/web/controller/InpatientVisitsController.java
mogoodrich: Yep, same thoughts above about @Resource and reps .  

For what it's worth, I said "slightly non-trivial", about the transfer part because the current query just looks for all active visits with an admit disposition without a subsequent "deny admission obs" and without any admission encounter.  Once you start dealing with transfers you need to support multiple transfers with a single visit... ie, I guess just getting all active visits where the most recent transfer disposition is before the most recent transfer encounter (?) so the query grows in complexity.

One thing I was thinking after I finished for the day on Monday was that although we want to support in the UI displaying both awaiting admission and awaiting transfer within the same list, they seem like fundamentally different things (and we will likely need to support different actions, ie admit vs transfer, in the UI) I was wondering if for clarify instead of expanding "getVisitsAwaitingAdmission" we create a new service method 
getVisitsAwaitingTransfer and the the O3 UI would call both and merge on the client... or we could have two API server methods but three endpoints, where the "getVisitsAwaitingAdmissionOrAwaitingTransfer" would query both methods and combine on the server side before shipping off to the client.

This might leave us with a more readable HQL query a well.

mogoodrich (commented):
Thanks @mseaton  makes sense, see my added thoughts!

mogoodrich (commented):
@mseaton made a few additions, namely:

* added an (unimplemented) getVisitsAwaitingTransfer service method
* added new endpoints to fetch transfers, and combined admission and transfers, setting a "type" property 
* Moved the time calculations into the VDW (and support the case where a patient is transferred without changing a ward)  (they should get some test methods, thought)

If we think this is a reasonable approach, I could potentially ticket the follow-on tasks.  (and start working on them, though likely not until July)

I did look a bit into support for parsing custom refs, but haven't come to a conclusion on that yet.  At minimum I believe we could take in a "rep" and parse it into the visit and patietn components and then rely on existing methods from there.

omod/src/main/java/org/openmrs/module/emrapi/web/controller/InpatientVisitsController.java
mseaton: I had similar thoughts @mogoodrich .  I think it probably makes sense as you describe to have have 2 separate backend  queries and 2 separate endpoints, and then consolidate the admission and transfer requests in the frontend.

api/src/main/java/org/openmrs/module/emrapi/visit/VisitDomainWrapper.java
```diff
@@ -532,6 +535,35 @@ public Location getInpatientLocation(Date onDate) {
...
 
+    public Integer timeSinceAdmissionInMinutes() {
+        if (getAdmissionEncounter() == null) {
+            return null;
+        } else {
+            return Minutes.minutesBetween(new DateTime(getAdmissionEncounter().getEncounterDatetime()), new DateTime()).getMinutes();
+        }
+    }
+
+    public Integer timeAtLocationInMinutes() {
```
mseaton, line 546: I'd rename this "timeAtInpatientLocationInMinutes".  This basically correspondes to the Location returned by the `getInpatientLocation` method

omod/src/main/java/org/openmrs/module/emrapi/web/controller/InpatientVisitsController.java
```diff
@@ -0,0 +1,106 @@
...
+@Controller
+@RequestMapping(value = "/rest/emrapi/inpatient")
+public class InpatientVisitsController {
+
+    @Autowired
+    private AdtService adtService;
+
+    @RequestMapping(method = RequestMethod.GET, value = "/visits")
+    @ResponseBody
+    public List<SimpleObject> getInpatientVisits(@RequestParam(value = "currentLocation") Location currentLocation) {
```
mseaton, line 30: We definitely need to improve this to _actually_ limit to the current inpatient location, not just return all visits at the parent visit location.  Not sure why this was implemented this way...

api/src/main/java/org/openmrs/module/emrapi/visit/VisitDomainWrapper.java
mogoodrich: Yep, make sense!

api/src/main/java/org/openmrs/module/emrapi/visit/VisitDomainWrapper.java
```diff
@@ -532,6 +535,35 @@ public Location getInpatientLocation(Date onDate) {
...
+        EncounterType admissionEncounterType = emrApiProperties.getAdmissionEncounterType();
+        EncounterType transferEncounterType = emrApiProperties.getTransferWithinHospitalEncounterType();
+
+        Location ward = null;
+        Integer time = null;
+
+        for (Encounter encounter : getSortedEncounters(SortOrder.MOST_RECENT_FIRST)) {
+            if (encounter.getEncounterType().equals(admissionEncounterType) || encounter.getEncounterType().equals(transferEncounterType)) {
+                time = Minutes.minutesBetween(new DateTime(encounter.getEncounterDatetime()), new DateTime()).getMinutes();
```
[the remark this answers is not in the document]
mogoodrich: I just did it like that to make sure I had something to test and fail against... :)  no, duh, will add the tests and fix...

mseaton (approved):

PR #233, opened 2024-07-09
EA-188  - Improve Java and Rest API for retrieving inpatient admission requests

mseaton:
See: https://openmrs.atlassian.net/browse/EA-188

Note, I have tested this fairly extensively against a production database backup that has several hundred active admission requests and related disposition data from the past 10 years, and many millions of observations overall to test performance.


mogoodrich (commented):
Generally looks really good, @mseaton , thanks!

My main way of reviewing was to to make sure that all the tests cases were there... the admit test cases look pretty robust (though per my comment on the code I think we should add some comment to clarify that this is only fetching "active" requests, though maybe that implicit in the name "request")?

Are you planning on adding tests for transfer and discharge, and particularly to confirm that "most recent disposition" always wins (I'm assuming that is what we want, at least, or we can discuss?)...  ie if a patient as a disposition of "Admit" during a course of a visit, but a subsequent disposition of "Discharge" (but with no admission encounter nor admission deny obs) the "Discharge" request should take precedent and the admission request should *not* be returned when requesting active admission requests?

api/src/main/resources/hql/inpatient_request_dispositions.hql
```diff
@@ -0,0 +1,42 @@
...
+    and (:visitLocation is null or visit.location = :visitLocation)
+    and person.dead = false
+    and visit.stopDatetime is null
+    and (
+        select count(*)
+        from Encounter as adtEncounter
+        where adtEncounter.visit = visit
+          and adtEncounter.voided = false
+          and adtEncounter.encounterType in (:adtEncounterTypes)
+    ) = 0
```
mogoodrich, line 23: Minor, but won't this exclude admission requests that come *after* an admission encounter (which is arguably just a edge/error case... it because more complicated if they are admitted, then discharged and then there's a new admission request within the same visit?  (though, again, it could be argued this should be a new visit?)

api/src/main/java/org/openmrs/module/emrapi/adt/InpatientRequestSearchCriteria.java
```diff
@@ -0,0 +1,33 @@
...
+import org.openmrs.module.emrapi.disposition.DispositionType;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Represents criteria for searching for AdtRequests
+ */
+@Data
+public class InpatientRequestSearchCriteria {
```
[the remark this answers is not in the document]
mseaton: Done.

api/src/main/resources/hql/inpatient_request_dispositions.hql
mseaton: This is another good point, and another case of this being copied from the pre-existing behavior, so the issue is long-standing.  I have added a unit test and a fix to both the old and new methods...have a look @mogoodrich

omod/src/main/java/org/openmrs/module/emrapi/web/controller/InpatientRequestController.java
```diff
@@ -0,0 +1,48 @@
...
+    @RequestMapping(method = RequestMethod.GET, value = "/rest/**/emrapi/inpatient/request")
+    @ResponseBody
+    public SimpleObject getInpatientRequests(
+            HttpServletRequest request,
+            HttpServletResponse response,
+            @RequestParam(required = false, value = "visitLocation") Location visitLocation,
+            @RequestParam(required = false, value = "dispositionLocation") List<Location> dispositionLocations,
+            @RequestParam(required = false, value = "dispositionType") List<DispositionType> dispositionTypes
+    ) {
+        RequestContext context = RestUtil.getRequestContext(request, response, Representation.REF);
```
mseaton, line 40: @mogoodrich here is another example of someplace I am just defaulting to REF.  How do we feel about this?

api/src/main/resources/hql/inpatient_request_dispositions.hql
mogoodrich: Yep, definitely realized this was pre-existing...

omod/src/main/java/org/openmrs/module/emrapi/web/controller/InpatientRequestController.java
mogoodrich: I guess I still think this is really the "default", not the "ref", but I don't feel super strongly.... you can ask Burke if you want? :)

mseaton: That's fine.  I've changed it to DEFAULT.  In the DB I'm testing on, on a request that returns 302 results, the `ref` representation is 458 KB uncompressed, and the `default` representation is 8.8 MB uncompressed, which is pretty significant.  Hopefully compression helps with that significantly, but generally using custom representations is likely what is called for here, especially as there is some duplication of properties (eg. patient + visit.patient)

api/src/main/java/org/openmrs/module/emrapi/adt/AdtServiceImpl.java
```diff
@@ -920,8 +924,85 @@ public List<Visit> getVisitsAwaitingAdmission(Location location, Collection<Inte
...
+        List<EncounterType> adtEncounterTypes = new ArrayList<>();
+        if (dispositionTypes.contains(DispositionType.ADMIT)) {
+            adtEncounterTypes.add(emrApiProperties.getAdmissionEncounterType());
+        }
+        if (dispositionTypes.contains(DispositionType.TRANSFER)) {
+            adtEncounterTypes.add(emrApiProperties.getTransferWithinHospitalEncounterType());
+        }
+        if (dispositionTypes.contains(DispositionType.DISCHARGE)) {
+            adtEncounterTypes.add(emrApiProperties.getExitFromInpatientEncounterType());
+        }
```
mseaton, line 964: This is kind of a question mark for me @mogoodrich .  The way I coded this up, if you ask for only ADMIT requests, then only Admission Encounters will be checked as those that might render them fulfilled.  Same with TRANSFER and DISCHARGE.  If we think that an admit disposition followed by a transfer or discharge encounter should lead to that admit request being fulfilled, then we may want to rethink this logic here, and just add all 3 encounter types in all cases.  Thoughts?

mogoodrich: More of a business logic thing but doubtful will get a answer, but my general thought is a request is "active" if 1) it is the most recent disposition in the visit, and 2) there hasn't been any successive "ADT encounter type" (so to your direct question, I would add all three encounter types) and 3) there is no successive "deny admission" obs (if the disposition is admit)

Thoughts? Am I missing anything?  At the end of the day, since a lot of the stuff is edge cases, probably best to just come up with the most succinct description/logic?  

FWIW we could come up with a better name than "active", but I think is better than "fulfilled"... ie in your above example (admit disposition followed by transfer or discharge), I don't think the request has been been "fulfilled", but it is no longer active, which is what we care about there?

mogoodrich (approved):

omod/src/main/java/org/openmrs/module/emrapi/web/controller/InpatientRequestController.java
mogoodrich: Not to re-open this, but to be clear, I think we could decide that the "default" rep of *this* endpoint only includes, for example, the "ref" reference to the underlying visit.  @mseaton

mogoodrich: I'm basically basing this on this pattern here: https://github.com/openmrs/openmrs-module-webservices.rest/blob/master/omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/PatientResource1_8.java#L121

mseaton: Fair.  So what is your opinion on what to return for the default and/or full representations?

mogoodrich: Full should return close to the full representation, within reason.  Default would return whatever we think is a "reasonable" default ref of this resource, balancing size vs useful information.  I know that's a vague answer but you are the one looking closest at what it actually returned now.  

For instance, taking a look at visit:

https://github.com/openmrs/openmrs-module-webservices.rest/blob/master/omod-1.9/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_9/VisitResource1_9.java

it would seem reasonable to return the DEFAULT visit rep by default (which looks like it generally only returns refs of nested resources) and the FULL visit rep if we ask for the FULL.

I haven't looked at the other properties... we might want just a REF version of the dispositionConcept since with all the names, etc, a default concept resource is likely overly verbose?  Do we anticipate needing anything other that the uuid of the disposition concept?

Anyway, to you point, we probably will need a custom rep at the end of the day, so I'd just pick something that is arguably "reasonable".   Hope that helps?

mseaton: OK, so I changed things so that when FULL is requested, you get a FULL visit, and a DEFAULT everything else, and when DEFAULT is requested you get a DEFAULT visit and a REF everything else.  Sound OK?

api/src/main/java/org/openmrs/module/emrapi/adt/AdtServiceImpl.java
mseaton: Ticketed separately here @mogoodrich :  https://openmrs.atlassian.net/browse/EA-195

omod/src/main/java/org/openmrs/module/emrapi/web/controller/InpatientRequestController.java
ibacher: That seems to make sense to me.

PR #234, opened 2024-07-11
EA-194: REST API url should have same standard prefix as other REST e…

mogoodrich:
…ndpoints in OpenMRS

https://openmrs.atlassian.net/browse/EA-194

### DIFF
--- a/omod/src/main/java/org/openmrs/module/emrapi/web/controller/EmrApiConfigurationController.java
+++ b/omod/src/main/java/org/openmrs/module/emrapi/web/controller/EmrApiConfigurationController.java
@@ -16,7 +16,7 @@ import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
 @Controller
-@RequestMapping(value = "/rest/emrapi/configuration")
+@RequestMapping(value = "/rest/**/emrapi/configuration")
 public class EmrApiConfigurationController {
 
     @Autowired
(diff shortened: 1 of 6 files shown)


omod/src/main/java/org/openmrs/module/emrapi/web/controller/EmrApiConfigurationController.java
```diff
@@ -16,7 +16,7 @@
 import javax.servlet.http.HttpServletResponse;
 
 @Controller
-@RequestMapping(value = "/rest/emrapi/configuration")
+@RequestMapping(value = "/rest/**/emrapi/configuration")
```
mseaton, line 19: This hasn't been released, so in theory we could just use `v1`, but this makes sense regardless I think until there is actually a v2+

omod/src/main/java/org/openmrs/module/emrapi/web/controller/InpatientVisitsController.java
```diff
@@ -19,7 +19,7 @@
 import org.springframework.web.bind.annotation.ResponseBody;
 
 @Controller
-@RequestMapping(value = "/rest/emrapi/inpatient")
+@RequestMapping(value = "/rest/**/emrapi/inpatient")
```
mseaton, line 22: Ditto.  Not really an issue, as I think this controller / endpoint is going to go away once my PR is ready...

mseaton (approved):
LGTM, thanks @mogoodrich

omod/src/main/java/org/openmrs/module/emrapi/web/controller/EmrApiConfigurationController.java
mogoodrich: Fair enough, I'm just going to merge this in, but feel free to change when you finalize your PR

mogoodrich:
Thanks @mseaton , yeah, feel free to rework the ones your actively working on

PR #235, opened 2024-07-22
EA-190 Improve API for retrieving inpatient visits

mseaton:
See https://openmrs.atlassian.net/browse/EA-190

This adds a new service method and Java components to facilitate finding and retrieving information about inpatient admissions, and exposes this RESTfully.


api/src/main/java/org/openmrs/module/emrapi/adt/AdtServiceImpl.java
```diff
@@ -1010,4 +1011,64 @@ public List<InpatientRequest> getInpatientRequests(InpatientRequestSearchCriteri
         }
         return ret;
     }
+
+    @Override
+    @Transactional(readOnly = true)
+    public List<InpatientAdmission> getInpatientAdmissions(InpatientAdmissionSearchCriteria criteria) {
+
```
mseaton, line 1018: This basically gets all ADT encounters associated with any active visit, and then groups these by visit into an InpatientAdmission object.

api/src/main/java/org/openmrs/module/emrapi/adt/InpatientAdmission.java
```diff
@@ -0,0 +1,105 @@
...
+import java.util.Comparator;
+import java.util.List;
+import java.util.Set;
+import java.util.TreeSet;
+
+/**
+ * Represents a hospital Admission
+ */
+@Data
+public class InpatientAdmission {
```
mseaton, line 20: Provides a mechanism to retrieve InpatientAdmission data and to expose various properties and utility getter methods via standard REST representations.

omod/src/main/java/org/openmrs/module/emrapi/rest/converter/InpatientAdmissionConverter.java
```diff
@@ -0,0 +1,44 @@
...
+@Handler(supports = InpatientAdmission.class, order = 0)
+public class InpatientAdmissionConverter extends SimpleBeanConverter<InpatientAdmission> {
+
+    private final Log log = LogFactory.getLog(getClass());
+
+    @Override
+    public DelegatingResourceDescription getResourceDescription(InpatientAdmission req, Representation representation) {
+        DelegatingResourceDescription ret = super.getResourceDescription(req, representation);
+        if (representation instanceof DefaultRepresentation) {
+            DelegatingResourceDescription rep = new DelegatingResourceDescription();
```
mseaton, line 22: This is a bit unconventional, but I went with this to focus on the data that is most relevant to the current use cases.  Specifically, this will enable getting `timeSinceAdmissionInMinutes`dynamically via `firstAdmissionOrTransferEncounter.encounterDatetime` and will enable getting `timeAtInpatientLocationInMinutes` dynamically via `encounterAssigningToCurrentInpatientLocation.encounterDatetime`

omod/src/main/java/org/openmrs/module/emrapi/web/controller/InpatientVisitsController.java
```diff
@@ -18,6 +18,7 @@
 
 @Controller
 @RequestMapping(value = "/rest/**/emrapi/inpatient")
+@Deprecated
```
mseaton, line 21: I have deprecated this for now, but would like to remove this controller altogether once we can get the ward app converted to the new endpoint.

omod/src/main/java/org/openmrs/module/emrapi/rest/converter/InpatientAdmissionConverter.java
mogoodrich: Makes sense.  Are we excluding the patient because a decent rep of the patient is included in the default rep of the visit?

omod/src/main/java/org/openmrs/module/emrapi/web/controller/InpatientVisitsController.java
mogoodrich: Fine with me

mogoodrich (approved):
thanks @mseaton !  added a couple comments but looks great to me, thanks for getting this out so quickly.

PR #237, opened 2024-08-05
EA-198: Add support for configuring Mother Child relationships within…

mogoodrich:
… the EMR API module


api/src/main/java/org/openmrs/module/emrapi/maternal/MaternalService.java
```diff
@@ -0,0 +1,23 @@
+package org.openmrs.module.emrapi.maternal;
+
+import java.util.List;
+
+import org.openmrs.Location;
+import org.openmrs.Patient;
+import org.openmrs.api.OpenmrsService;
+
+public interface MaternalService extends OpenmrsService {
```
mogoodrich, line 9: Better ideas for a name of this? MaternityService? MaternalCareService?  Loop it into another service in EMR API (it doesn't really seem like an ADT method?)

(Note that there will like be a getMotherForNewborn()) service method as well.

api/src/main/java/org/openmrs/module/emrapi/maternal/Newborn.java
```diff
@@ -0,0 +1,11 @@
+package org.openmrs.module.emrapi.maternal;
+
+import lombok.Data;
+import org.openmrs.Patient;
+import org.openmrs.Visit;
+
+@Data
+public class Newborn {
+    private Patient newborn;
+    private Visit newbornVisit;
```
mogoodrich, line 10: We can evolve what this object should contain... it likely needs to contain admission information, but I haven't checked out the latest InpatientAdmission / InpatientRequest data structure and I wanted to see how that was evolving before digging too deep into this.

mogoodrich (commented):
I left this in draft format because the omod REST endpoint stuff still needs to be added but I wanted to put this out for review sooner rather than later, especially to get feedback on the name of the new service and overall structure, see my comments.
@cioan  @mseaton  @chibongho  flagging you all for review.

api/src/main/java/org/openmrs/module/emrapi/maternal/MaternalService.java
mseaton: You might just be able to slot this into the AdtService, since you are generally dealing with admitted patients here.  It would simplify the commit.  Otherwise, could potentially add to EmrPatientService.  Or, start some new, more generic service that we can add to when needed.

api/src/main/java/org/openmrs/module/emrapi/EmrApiConstants.java
```diff
@@ -215,4 +215,6 @@ public class EmrApiConstants {
     ));
 
     public static final String GP_USE_LEGACY_DIAGNOSIS_SERVICE = "emrapi.useLegacyDiagnosisService";
+
+    public static final String METADATA_MAPPING_MOTHER_CHILD_RELATIONSHIP_TYPE = "emrapi.motherChildRelationshipType";
```
mseaton, line 219: Do you think we need to support A/B vs. B/A variations?

api/src/main/java/org/openmrs/module/emrapi/maternal/MaternalServiceImpl.java
```diff
@@ -0,0 +1,62 @@
...
+
+    public void setEmrApiProperties(EmrApiProperties emrApiProperties) {
+        this.emrApiProperties = emrApiProperties;
+    }
+
+    public void setEmrApiDAO(EmrApiDAO emrApiDAO) {
+        this.emrApiDAO = emrApiDAO;
+    }
+
+    public List<Newborn> getNewbornsByMother(Patient mother, Location visitLocation) {
```
mseaton, line 31: I'm a little worried about the efficiency of this.  I agree we want to support being able to getNewborns for a single mother, but we may need the overall method to be more efficient to operate in bulk in some way (eg. take in a List of mother patient ids and get back a Map<Mother, List<Baby>> or something.

api/src/main/java/org/openmrs/module/emrapi/maternal/MaternalServiceImpl.java
```diff
@@ -0,0 +1,62 @@
...
+        }
+
+        if (mother == null) {
+            throw new APIException("Mother cannot be null");
+        }
+
+        Map<String, Object> parameters = new HashMap<>();
+        parameters.put("mother", mother);
+        parameters.put("motherChildRelationshipType", motherChildRelationshipType);
+        parameters.put("visitLocation", visitLocation);
```
mseaton, line 46: What is the purpose of the visitLocation parameter here?  If "mother" is required, are you saying "but don't return any babies for the mother if the visitLocation doesn't match"?  No, you should already have filtered by visitLocation to get the valid mothers.

api/src/main/resources/hql/newborns_by_mother.hql
```diff
@@ -0,0 +1,24 @@
+select
+    baby,
+    babyVisit
+from
+    Person as mother,
+    Person as baby,
+    Relationship as motherChildRelationship,
+    Visit as motherVisit,
+    Visit as babyVisit
```
mseaton, line 9: It would be more consistent with the other hql I recently added to connect the tables via joins rather than where criteria.

api/src/main/resources/hql/newborns_by_mother.hql
```diff
@@ -0,0 +1,24 @@
...
+where
+    mother = :mother
+    and motherChildRelationship.personA = mother
+    and motherChildRelationship.personB = baby
```
mseaton, line 13: Per my comment above, do we need to support the inverse here?

api/src/main/resources/hql/newborns_by_mother.hql
```diff
@@ -0,0 +1,24 @@
...
+    and motherChildRelationship.relationshipType = :motherChildRelationshipType
+    and motherVisit.patient = mother and motherVisit.stopDatetime is null
+    and babyVisit.patient = baby and babyVisit.stopDatetime is null
+    and year(baby.birthdate) >= year(motherVisit.startDatetime)
+    and month(baby.birthdate) >= month(motherVisit.startDatetime)
+    and day(baby.birthdate) >= day(motherVisit.startDatetime)
```
mseaton, line 19: are year/month/day like this more efficient than just casting to date? `and baby.birthdate >= date(motherVisit.startDatetime)`

api/src/main/resources/hql/newborns_by_mother.hql
```diff
@@ -0,0 +1,24 @@
...
+    and (:visitLocation is null or (motherVisit.location = :visitLocation and babyVisit.location = :visitLocation))
```
mseaton, line 20: I really don't see us ever limiting by visitLocation.  Even in the existing adt queries, I don't think we really have that use case...

api/src/main/java/org/openmrs/module/emrapi/EmrApiConstants.java
mogoodrich: No.  I mean, at least not yet.  I think it's fair to say "to use the newborn functionality you will need to have a "mother-child" relationship type, where a_is_to_b=Mother and b_is_to_a=Child (which I think is implied in the name "mother-child).  If we discover that there's an existing implementation with a "Child-Mother relationship type" that wants to use this functionality we can consider supporting the B/A variation.  But I think by supporting both off the bat we are potentially encouraging/enabled implementations to not standardize.

api/src/main/java/org/openmrs/module/emrapi/maternal/MaternalService.java
mogoodrich: I'm fine with putting this in ADTService, or moving it to the EmtrPatientService.    I don't really want to create a new generic service because we will just be guessing on a good generic name and I suspect we will regret whatever name we choose down the road.

api/src/main/java/org/openmrs/module/emrapi/maternal/MaternalServiceImpl.java
mogoodrich: Fair.  It's likely easy enough to support multiple patients.  Though I'm a little worried about how we'd hand this on the client side.  ie, it's nice to be able to separate the concerns of how to fetch the relevant data needed to display the patient card from fetching the list of the patients on the ward.

api/src/main/java/org/openmrs/module/emrapi/maternal/MaternalServiceImpl.java
mogoodrich: Good point.  I'll sleep on this in case I'm forgetting something, but that does seem to make sense.

api/src/main/resources/hql/newborns_by_mother.hql
mogoodrich: Would have much preferred to use "date" but as far as I could tell it wasn't available in HQL

api/src/main/java/org/openmrs/module/emrapi/EmrApiConstants.java
mseaton: Makes sense, agreed.

api/src/main/java/org/openmrs/module/emrapi/maternal/MaternalService.java
mseaton: OK, defer to you.

api/src/main/java/org/openmrs/module/emrapi/maternal/MaternalServiceImpl.java
mseaton: Yeah, agree, just wanted to flag it.

api/src/main/resources/hql/newborns_by_mother.hql
mseaton: Oh right.  Hql.

api/src/main/java/org/openmrs/module/emrapi/maternal/MaternalServiceImpl.java
mogoodrich: I think it's easy enough to support 1 to n, I will likely add that.

api/src/main/java/org/openmrs/module/emrapi/maternal/MaternalService.java
mogoodrich: So, I started to move this into AdtService, but I like the ideal of smaller, targeted services rather than bloat, so I'd prefer to keep this separate for "Maternal" functionality, especially now that we have two methods.

api/src/main/java/org/openmrs/module/emrapi/maternal/MaternalService.java
```diff
@@ -0,0 +1,30 @@
...
+
+    /**
+     * Returns all mothers of the specified patients(), where "mother" is defined as a patient who is:
+     * - linked to the specified patient by a relationship of typ emrapi.motherChildRelationshipType
+     * - has an active visit at the same visit location of the newborn
+     *
+     * @param newborns  (assumption: these are newborns, method does *not* confirm this)
+     * @return
+     */
+    public List<Mother> getMothersByNewborn(List<Patient> newborns);
```
mogoodrich, line 29: Added this since the last commit.

One thing it *doesn't* do is confirm that the newborns are actually "newborns", so the client side will have to handle this; easy enough to support this if we want to, but it seemed weird to have the method potentially one of the inputs.  Practically, this would only make a different if there were a mother and a child in the same hospital and the child is *not*  a newborn... the mother would still show up on the child's card, which certainly isn't the end of the world.

I also considered a single method that, given a patient set, would use a single query to fetch both NewbornsByMother and MothersByNewborn.  This seemed like it would be straightfoward to write and would save us a REST call and a query, but it seemed like things would be getting too confusing for the consumer to understand.

api/src/main/java/org/openmrs/module/emrapi/maternal/MaternalServiceImpl.java
```diff
@@ -0,0 +1,134 @@
...
+        for (Object req : l) {
+            Object[] row = (Object[]) req;
+            Newborn newborn = new Newborn();
+            newborn.setNewborn((Patient) row[0]);
+            newborn.setMother((Patient) row[1]);
+            visits.add((Visit) row[2]);
+            ret.add(newborn);
+        }
+
+        // now fetch all the admissions for newborns in the result set
```
mogoodrich, line 71: We need the inpatient admission for the newborn so we can should their current location.  I followed a similar pattern to what @mseaton  for fetching all inpatient requests associated with inpatient admission.

api/src/main/resources/hql/mothers_by_newborn.hql
```diff
@@ -0,0 +1,18 @@
...
+from
+    Relationship as motherChildRelationship, Visit as motherVisit, Visit as babyVisit
+    inner join motherChildRelationship.personA as mother
+    inner join motherChildRelationship.personB as baby
+where
+  baby in (:babies)
+  and motherChildRelationship.relationshipType = :motherChildRelationshipType
+  and motherVisit.patient = mother and motherVisit.stopDatetime is null
+  and babyVisit.patient = baby and babyVisit.stopDatetime is null
+  and babyVisit.location = motherVisit.location
```
mogoodrich, line 14: I added this restriction (mother and baby visits must be at the same location). This might be problematic for implementations that collect visit location at a greater level of granularity than at the facility level, but it seemed like something we should do in preparation for a more cloud-based model where there are multiple facilities in the same database.

mogoodrich:
Moving this out of draft state even though it still doesn't have the code to expose these service methods as REST endpoints, will likely work on that as a separate PR.

I have made changes and added a baby-for-mother inverse service method and added some comments, flagging @mseaton  for re-review.

api/src/main/resources/hql/newborns_by_mother.hql
```diff
@@ -0,0 +1,21 @@
+select
+    baby,
+    mother,
+    babyVisit
+from
+    Relationship as motherChildRelationship, Visit as motherVisit, Visit as babyVisit
+    inner join motherChildRelationship.personA as mother
+    inner join motherChildRelationship.personB as baby
```
mogoodrich, line 8: And, oh, I converted this somewhat to inner joins, but needed help on how do this when can't just chain the joins... ie relationship links to person/patient, but then visit links to patient not the other way around... hope that made some sense and I'm not just missing something obvious.  :)

api/src/main/java/org/openmrs/module/emrapi/maternal/MaternalService.java
```diff
@@ -0,0 +1,30 @@
...
+    /**
+     * Returns all "newborns" of the specified patients, where "newborn" is defined as a patient who is:
+     *  - linked to the specified patient by a relationship of type emrapi.motherChildRelationshipType
+     *  - has as an active visit at the same visit location of the mother
+     *  - has a birthdate that is on or after the start date of the mother's active visit (at the visitLocation, if specified) (note matches on date, not datetime to account for retrospective data entry or only have a date component of birthdate)
+     *
+     * @param mothers
+     * @return
+     */
+    public List<Newborn> getNewbornsByMother(List<Patient> mothers);
```
mseaton, line 19: This is where I found a lot of value (and really like the pattern) of using `SearchCriteria`.  So if we continue with these methods (and I have other feedback that may supercede this), it might be preferable to do something like `getNewborns(NewbornSearchCriteria)`, and then one of the properties of `NewbornSearchCriteria` might be which mothers to search on, but other use cases can emerge to get Newborns by something other than their mothers (or in addition).

Also, it is potentially more expensive to take in a List<Patient> here, if you need to look up a bunch of patient by id or uuid before passing them in.

api/src/main/java/org/openmrs/module/emrapi/maternal/MaternalService.java
mseaton: Same comment as above.  `getMothers(MotherSearchCriteria)`.  But see my other comments below also.

api/src/main/java/org/openmrs/module/emrapi/maternal/MaternalServiceImpl.java
mseaton: So Mother and Newborn are inherently _admitted_ patients who just delivered / were born within the facility, but I'm not sure that is necessarily obvious to consumers of a MaternalService that are getting mothers and babies.

I think I might just move everything into the `adt` domain and `AdtService` if we are dealing with ADT things - in this case, these are really more specific instances of InpatientAdmission (eg. subclass or wrapper class).

I think it would be more intuitive if `Mother` -> `MaternalAdmission` and `Newborn` -> `NewbornAdmission`, both of which `extend InpatientAdmission`.  

Then we add methods to AdtService for `getMaternalAdmissions(MaternalAdmissionSearchCriteria)` and `getNewbornAdmissions(NewbornAdmissionSearchCriteria)`

```
  MaternalAdmission extends InpatientAdmission (
    private List<NewbornAdmission> newbornAdmissions;
  )
```

```
  NewbornAdmission extends InpatientAdmission (
    private List<MaternalAdmission> maternalAdmission;
  )
```

Then the ward app can just get these directly, without making any additional queries...

api/src/main/java/org/openmrs/module/emrapi/maternal/Mother.java
```diff
@@ -0,0 +1,13 @@
...
+import lombok.Data;
+import org.openmrs.Patient;
+import org.openmrs.module.emrapi.adt.InpatientAdmission;
+
+
+@Data
+public class Mother {
+    private Patient mother;
+    private Patient newborn;
+    private InpatientAdmission motherAdmission;
```
mseaton, line 12: See comment above where we could make `Mother` -> `MaternalAdmission`

api/src/main/java/org/openmrs/module/emrapi/maternal/Newborn.java
```diff
@@ -0,0 +1,12 @@
...
+
+import lombok.Data;
+import org.openmrs.Patient;
+import org.openmrs.module.emrapi.adt.InpatientAdmission;
+
+@Data
+public class Newborn {
+    private Patient newborn;
+    private Patient mother;
+    private InpatientAdmission newbornAdmission;
```
mseaton, line 11: See comment above where we could make `Newborn` -> `NewbornAdmission`

api/src/main/resources/hql/mothers_by_newborn.hql
mseaton: I feel like this is more likely to cause data to be excluded that shouldn't be, than include too much data, at least in the short term.  We can always add this restriction in, but I'd be a little wary about it.  I don't think there are a lot of use cases where we want to exclude babies from mothers who were born after the mothers visit start date, because they are in different facilities in a federated system.  Seems unlikely.

api/src/main/resources/hql/mothers_by_newborn.hql
```diff
@@ -0,0 +1,18 @@
...
+    motherVisit
+from
+    Relationship as motherChildRelationship, Visit as motherVisit, Visit as babyVisit
+    inner join motherChildRelationship.personA as mother
+    inner join motherChildRelationship.personB as baby
+where
+  baby in (:babies)
+  and motherChildRelationship.relationshipType = :motherChildRelationshipType
+  and motherVisit.patient = mother and motherVisit.stopDatetime is null
+  and babyVisit.patient = baby and babyVisit.stopDatetime is null
```
mseaton, line 13: Maybe this is going over the same ground as my other comments, but I don't love all of these restrictions here in a "generic" mothers by newborn method.  If someone were to see a getMothersByNewborn method, passing in a Newborn, and got null back for the mother because the mother no longer had an active visit, that would seem...unexpected.  Another reason why I like being more explicit with our naming around XxxAdmission

api/src/main/resources/hql/newborns_by_mother.hql
```diff
@@ -0,0 +1,21 @@
...
+where
+    mother in (:mothers)
+    and motherChildRelationship.relationshipType = :motherChildRelationshipType
+    and motherVisit.patient = mother and motherVisit.stopDatetime is null
+    and babyVisit.patient = baby and babyVisit.stopDatetime is null
```
mseaton, line 13: Same comment as before.  Consumers will wonder why no babies are returned for a mother just because they no longer have an active visit.  That seems more like a front-end concern than the concern of a back-end service method, unless we make this much more explicitly specific.

mseaton (changes requested):

api/src/main/java/org/openmrs/module/emrapi/maternal/MaternalService.java
mogoodrich: Passing in uuids/ids likely makes sense here, good point...

mogoodrich (commented):
Okay, ready for re-review @mseaton ... I still need to add the REST endpoint and add new javadocs for thee new service but this updated commit now:

* Uses search criteria objects for Mother and Child searches, enabling configuring whether to limit to patients with actvive visits
* Uses a list of patient uuids for the search instead of patient objects
* Standardizing naming convention on child (instead of "baby" and "newborn")

api/src/main/java/org/openmrs/module/emrapi/maternal/Child.java
```diff
@@ -0,0 +1,12 @@
+package org.openmrs.module.emrapi.maternal;
+
+import lombok.Data;
+import org.openmrs.Patient;
+import org.openmrs.module.emrapi.adt.InpatientAdmission;
+
+@Data
+public class Child {
+    private Patient child;
+    private Patient mother;
```
mseaton, line 10: Thoughts about making this to be `private Mother mother;`?

api/src/main/java/org/openmrs/module/emrapi/maternal/Mother.java
```diff
@@ -0,0 +1,13 @@
...
+
+import lombok.Data;
+import org.openmrs.Patient;
+import org.openmrs.module.emrapi.adt.InpatientAdmission;
+
+
+@Data
+public class Mother {
+    private Patient mother;
+    private Patient child;
```
mseaton, line 11: Thoughts about making this `private Child child;`?

api/src/main/java/org/openmrs/module/emrapi/maternal/Child.java
mogoodrich: I'd rather keep the REST response as simple as possible.  You mean have:

```
public class Mother {
    private Patient mother;
    private Child child;
    private InpatientAdmission motherAdmission;
}


public class Child {
    private Patient child;
    private Mother mother;
    private InpatientAdmission childAdmission;
}
```

Then on a child you've have to child.mother.mother to get the mother?

api/src/main/java/org/openmrs/module/emrapi/maternal/Mother.java
mogoodrich: See above

api/src/main/java/org/openmrs/module/emrapi/maternal/MotherSearchCriteria.java
```diff
@@ -0,0 +1,22 @@
+package org.openmrs.module.emrapi.maternal;
+
+import java.util.List;
+
+import lombok.Data;
+
+@Data
```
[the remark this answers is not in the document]
mogoodrich: Done

api/src/main/java/org/openmrs/module/emrapi/maternal/MaternalService.java
```diff
@@ -0,0 +1,14 @@
...
+
+import java.util.List;
+
+import org.openmrs.Patient;
+import org.openmrs.api.OpenmrsService;
+
+public interface MaternalService extends OpenmrsService {
+
+    // TODO update javadoc
+    public List<Child> getChildrenByMother(ChildSearchCriteria criteria);
```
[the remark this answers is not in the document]
mogoodrich: So I changed this to getChildren, but (almost) back to the original, making the mothers plural: "getChildrenByMothers".  I *did* remove making setting mother uuids mandatroy, but I think the name should stay, reflecting the case that this returns child-mother pairs, (and, in particular, would return two rows for single child if they had two or mother mothers... admittedly, not a standard case, but certainly a standard case for the inverse, getMothersByChild).

api/src/main/java/org/openmrs/module/emrapi/maternal/MaternalService.java
```diff
@@ -0,0 +1,14 @@
...
+
+    public List<Mother> getMothersByChild(MotherSearchCriteria criteria);
```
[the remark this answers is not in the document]
mogoodrich: See above (and I did remove making setting child uuids mandatory)

api/src/main/resources/hql/mothers_by_child.hql
```diff
@@ -0,0 +1,16 @@
...
+    inner join motherChildRelationship.personA as mother
+    inner join motherChildRelationship.personB as child
+where
+  child.uuid in (:childUuids)
+  and motherChildRelationship.relationshipType = :motherChildRelationshipType
+  and (:motherHasActiveVisit = false or (select count(motherVisit) from Visit as motherVisit where motherVisit.patient = mother and motherVisit.stopDatetime is null and motherVisit.voided = false) > 0)
+  and (:childHasActiveVisit = false or (select count(childVisit) from Visit as childVisit where childVisit.patient = child and childVisit.stopDatetime is null and childVisit.voided = false) > 0)
+  and mother.voided = false and child.voided = false and motherChildRelationship.voided = false
+
+
```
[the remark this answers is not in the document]
mogoodrich: Done.  And now both support childBornDuringMothersActiveVisit

mogoodrich (commented):
@mseaton  I made most, but not all of the suggested changes, see my comments and updated PR.

Going to go ahead and merge this in.

api/src/main/java/org/openmrs/module/emrapi/maternal/ChildrenByMothersSearchCriteria.java
```diff
@@ -0,0 +1,20 @@
...
+import lombok.experimental.Accessors;
+
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+public class ChildrenByMothersSearchCriteria {
+    private List<String> motherUuids;  // restrict to children of these mothers
+    @Accessors(fluent = true)  private boolean requireMotherHasActiveVisit = false;  // restrict to children of mothers who have an active visit
+    @Accessors(fluent = true)  private boolean requireChildHasActiveVisit = false;   // restrict to children who have an active visit
+    @Accessors(fluent = true)  private boolean requireChildBornDuringMothersActiveVisit = false;  // restrict to children who were born during their mother's active visit
```
mseaton, line 17: What is this `fluent = true`?  Don't we want these to act and behave like standard Javabean properties?

api/src/main/java/org/openmrs/module/emrapi/maternal/Child.java
mseaton: Hmm.  I think given what you have in place it would make sense to just have one object:

```
public class MotherAndChild {
    private Patient mother;
    private Child child;
    private InpatientAdmission motherAdmission;
    private InpatientAdmission childAdmission;
}
```

And then one service method to `getMothersAndChildren(MothersAndChildrenSearchCriteria)` that returns a List of these.

That list would contain all children of interest if you pass it a set of child uuids, and would contain all mothers of interest if you pass it was set of mother uuids, and everyone of interest if you just pass it criteria about the visit constraints expected...

api/src/main/java/org/openmrs/module/emrapi/maternal/ChildrenByMothersSearchCriteria.java
mogoodrich: It makes the getter/setter "requireMotherHasActiveVisit" instead of "isRequireMotherHasActiveVisit"... "is" is the default for "boolean" types and that read horribly. https://projectlombok.org/features/experimental/Accessors

mogoodrich: No idea why it's called fluent. :)

api/src/main/java/org/openmrs/module/emrapi/maternal/MaternalService.java
```diff
@@ -0,0 +1,22 @@
...
+     * @return a list of children, with their linked mothers (note this returns a "Child" per mother-child pair, so a child with multiple mothers will appear multiple times)
+     */
+    List<Child> getChildrenByMothers(ChildrenByMothersSearchCriteria criteria);
+
+    /**
+     * Fetches patients who are "mothers" (personA) of a Mother-Child relationship
+     * @param criteria search criteria (see class for details)
+     * @return a list of mothers, with their linked children (note this returns a "Mother" per mother-child pair, so a mother with multiple children will appear multiple times)
+     */
+    List<Mother> getMothersByChildren(MothersByChildrenSearchCriteria criteria);
```
mseaton, line 21: Regarding these method names, see my suggestion above.  These aren't "ByMothers" or "ByChildren" at all.  They are really lists of MotherAndChild objects.  They aren't "by" anything.

api/src/main/java/org/openmrs/module/emrapi/maternal/Mother.java
mseaton: See above again.

api/src/main/java/org/openmrs/module/emrapi/maternal/MothersByChildrenSearchCriteria.java
```diff
@@ -0,0 +1,19 @@
...
+import lombok.experimental.Accessors;
+
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+public class MothersByChildrenSearchCriteria {
+    private List<String> childUuids;  // restrict to mothers of these children
+    @Accessors(fluent = true) private Boolean requireMotherHasActiveVisit = false;  // restrict to mothers who have an active visit
+    @Accessors(fluent = true) private Boolean requireChildHasActiveVisit = false;  // restrict to mothers of children who have an active visit
+    @Accessors(fluent = true)  private boolean requireChildBornDuringMothersActiveVisit = false;  // restrict to mothers who had a child born during their active visit
```
mseaton, line 17: Ditto

api/src/main/resources/hql/mother_child.hql
```diff
@@ -0,0 +1,18 @@
+select
+    mother,
+    child
+from
+    Relationship as motherChildRelationship
+        inner join motherChildRelationship.personA as mother
+        inner join motherChildRelationship.personB as child
```
ibacher, line 7: Are relationships always uni-directional like this, i.e., with a fixed A/B? (The flexibility around this is one of the things I've always disliked about relationships).

api/src/main/java/org/openmrs/module/emrapi/maternal/ChildrenByMothersSearchCriteria.java
mseaton: I understand what it does, but nothing else that works with JavaBean objects will be able to deal with this.  We should make it standard, even if it reads poorly, or try to change the variable name so it reads better with an "is" or "get" prefix.

api/src/main/java/org/openmrs/module/emrapi/maternal/Child.java
mogoodrich: Yeah, I was thinking of that, though I thought there was a reason it wasn't going to work, will check again.

api/src/main/resources/hql/mother_child.hql
mogoodrich: They are flexible, but we are saying this is based around have a standardized uni-directional "Mother to Child" relationship type

api/src/main/java/org/openmrs/module/emrapi/maternal/ChildrenByMothersSearchCriteria.java
mogoodrich: ah... let's just stick with the ugly name then, will change.

PR #238, opened 2024-08-08
EA-198-2: Add support for configuring (and fetching) Mother Child rel…

mogoodrich:
…ationships within the EMR API module


mogoodrich:
Okay, took a while to rename all the classes and variables in the test case, but, mircalously all the tests passed first time after it compiled.  I probably need to give this one more once-over before merging, and add a couple tests around the weird cases where you are restricting by both mother and child in a single request.  Will take a look tomorrow, but wanted to get this open for review.

api/src/main/java/org/openmrs/module/emrapi/maternal/MaternalServiceImpl.java
```diff
@@ -46,79 +47,40 @@ public List<Child> getChildrenByMothers(ChildrenByMothersSearchCriteria criteria
...
 
         List<?> l = emrApiDAO.executeHqlFromResource("hql/mother_child.hql", parameters, List.class);
 
-        List<Child> ret = new ArrayList<>();
+        List<MotherAndChild> ret = new ArrayList<>();
 
         for (Object req : l) {
             Object[] row = (Object[]) req;
-            Child child = new Child();
+            MotherAndChild child = new MotherAndChild();
```
mseaton, line 62: Probably should rename this local variable to `motherAndChild`, though you could also just have a constructor that takes in both and not need a local variable at all...

mseaton (approved):
A few minor comments, but I think this looks like a good improvement, simplifies things and also makes things cleaner and clearer

api/src/main/java/org/openmrs/module/emrapi/maternal/MaternalServiceImpl.java
mogoodrich: Whoops, yep... missed this one.

api/src/main/java/org/openmrs/module/emrapi/maternal/MothersAndChildrenSearchCriteria.java
```diff
@@ -0,0 +1,19 @@
...
+
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+public class MothersAndChildrenSearchCriteria {
+    private List<String> motherUuids;  // restrict to children of these mothers
+    private List<String> childUuids;  // restrict to mothers of these children
+    private boolean requireMotherHasActiveVisit = false;  // restrict to mothers who have an active visit
+    private boolean requireChildHasActiveVisit = false;  // restrict to mothers of children who have an active visit
+    private boolean requireChildBornDuringMothersActiveVisit = false;  // restrict to mothers who had a child born during their active visit
```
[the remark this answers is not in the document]
mogoodrich: I like that.

PR #240, opened 2024-08-13
TRUNK-6188: Add whitelisting for components loaded via XStream

ibacher:
Adds whitelist entries for the MetadataPackagesConfig classes supported by this module.


mseaton (approved):
Seems ok.

PR #241, opened 2024-09-05
EA-201 - Add additional method to AccountService to find accounts by …

mseaton:
…name


mogoodrich (approved):
LGTM, thanks @mseaton !

PR #243, opened 2024-09-06
EA-195 - ADT inpatient requests should be fulfilled by any ADT encounter

mseaton:
See https://openmrs.atlassian.net/browse/EA-195


mogoodrich (commented):
LGTM!

PR #251, opened 2024-10-31
EA-208: Change Inpatient Admission and Inaptient Request endpoints to…

chibongho:
… support searching by patient and visits

Ticket: https://openmrs.atlassian.net/browse/EA-208

### DIFF
--- a/omod/src/main/java/org/openmrs/module/emrapi/web/controller/InpatientRequestController.java
+++ b/omod/src/main/java/org/openmrs/module/emrapi/web/controller/InpatientRequestController.java
@@ -1,6 +1,8 @@
 package org.openmrs.module.emrapi.web.controller;
 
 import org.openmrs.Location;
+import org.openmrs.Patient;
+import org.openmrs.Visit;
 import org.openmrs.module.emrapi.adt.AdtService;
 import org.openmrs.module.emrapi.adt.InpatientRequest;
 import org.openmrs.module.emrapi.adt.InpatientRequestSearchCriteria;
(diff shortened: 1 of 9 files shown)


omod/src/main/java/org/openmrs/module/emrapi/web/controller/InpatientRequestController.java
```diff
@@ -35,13 +35,17 @@ public SimpleObject getInpatientRequests(
...
+            @RequestParam(required = false, value = "dispositionType") List<DispositionType> dispositionTypes,
+            @RequestParam(required = false, value = "patients") List<String> patients,
+            @RequestParam(required = false, value = "visits") List<String> visits
     ) {
         RequestContext context = RestUtil.getRequestContext(request, response, Representation.DEFAULT);
         InpatientRequestSearchCriteria criteria = new InpatientRequestSearchCriteria();
         criteria.setVisitLocation(visitLocation);
         criteria.setDispositionLocations(dispositionLocations);
         criteria.setDispositionTypes(dispositionTypes);
+        criteria.setPatients(patients);
```
mogoodrich, line 47: Can't we just map patients (and visits) to their ids here and keep the criteria the same?

mogoodrich (changes requested):
Yeah, I agree with @mseaton ... and I think it should just be as simple as mapping the incoming patients and visits to their underlying ids (available via getId()) and passing them into the existing criteria?  

Thanks @chibongho !

mogoodrich (approved):
LGTM, thanks @chibongho !

PR #252, opened 2024-11-07
EA-209 fix date comparison in getMothersAndChildren() to determine wh…

chibongho:
…ether child is born during mother's visit

See: https://openmrs.atlassian.net/browse/EA-209

### DIFF
--- a/api/src/main/resources/hql/mother_child.hql
+++ b/api/src/main/resources/hql/mother_child.hql
@@ -12,7 +12,6 @@ where
   and (:motherRequiredToHaveActiveVisit = false or (select count(motherVisit) from Visit as motherVisit where motherVisit.patient = mother and motherVisit.stopDatetime is null and motherVisit.voided = false) > 0)
   and (:childRequiredToHaveActiveVisit = false or (select count(childVisit) from Visit as childVisit where childVisit.patient = child and childVisit.stopDatetime is null and childVisit.voided = false) > 0)
   and (:childRequiredToBeBornDuringMothersActiveVisit = false or (select count(motherVisit) from Visit as motherVisit where motherVisit.patient = mother and motherVisit.stopDatetime is null and motherVisit.voided = false
-                                                                                                                   and year(child.birthdate) >= year(motherVisit.startDatetime)
-                                                                                                                   and month(child.birthdate) >= month(motherVisit.startDatetime)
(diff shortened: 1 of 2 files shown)


api/src/main/resources/hql/mother_child.hql
```diff
@@ -12,7 +12,6 @@ where
...
   and (:childRequiredToHaveActiveVisit = false or (select count(childVisit) from Visit as childVisit where childVisit.patient = child and childVisit.stopDatetime is null and childVisit.voided = false) > 0)
   and (:childRequiredToBeBornDuringMothersActiveVisit = false or (select count(motherVisit) from Visit as motherVisit where motherVisit.patient = mother and motherVisit.stopDatetime is null and motherVisit.voided = false
-                                                                                                                   and year(child.birthdate) >= year(motherVisit.startDatetime)
-                                                                                                                   and month(child.birthdate) >= month(motherVisit.startDatetime)
-                                                                                                                   and day(child.birthdate) >= day(motherVisit.startDatetime)) > 0)
+                                                                                                                   and (year(child.birthdate)||'-'||month(child.birthdate)||'-'||day(child.birthdate)) >=
```
chibongho, line 15: I don't like this manual concatenation, but I can't find a `DATE()` function in hql like in mysql.

mseaton (commented):
Ick, ugly, but it was also ugly before.

chibongho:
A bit moot, but looks like later version of hql has something saner:

    extract(date from child.birthdate)

https://docs.jboss.org/hibernate/orm/6.6/userguide/html_single/Hibernate_User_Guide.html

mogoodrich:
Oh, thanks @chibongho , yeah, whoops, bug on my part, nice catch.  The string concanetation is ugly, but if we are assured that the greater than test will always work correctly, I guess we are okay.

The other option would be to be more explicit about the day/month/year comparisions, we could do this, right (admittedly "wordy", and I probably have my parathesis wrong, hard to match up outside of an editor)?

```
and (year(child.birthdate) >= year(motherVisit.startDatetime) or 

(year(child.birthdate) == year(motherVisit.startDatetime) and month(child.birthdate) >= month(motherVisit.startDatetime)) or

(year(child.birthdate) == year(motherVisit.startDatetime) and month(child.birthdate) == month(motherVisit.startDatetime)) and day(child.birthdate) >= day(motherVisit.startDatetime)) 
```

PR #255, opened 2024-11-12
RA-2030: Fix UI framework error in the reference application

wikumChamith:


wikumChamith:
@dkayiwa wouldn't it be safer to rename this rather than remove it?

api/src/main/java/org/openmrs/module/emrapi/EmrApiActivator.java
```diff
@@ -73,8 +73,6 @@ public class EmrApiActivator extends BaseModuleActivator implements DaemonTokenA
 
     private PersonService personService;
 
-    private ConceptService conceptService;
```
dkayiwa, line 76: Why are you removing this?

wikumChamith: This variable does not get used after removing `createConceptSource(conceptService);`

wikumChamith: It is what creates the duplicate `org.openmrs.module.emrapi` concept_reference_source. According to this comment this method was created for use in tests.


https://github.com/openmrs/openmrs-module-emrapi/blob/35032fae6eb22b6be044f8000fcaf4144b846eaf/api/src/main/java/org/openmrs/module/emrapi/EmrApiActivator.java#L343-L358

dkayiwa: So where are we creating the original org.openmrs.module.emrapi concept source from?

mogoodrich: Not sure why that comment is still there, but, yeah, it looks like this is what creates the org.openmrs.module.emrapi concept source and is needed?  It looks like the method *should* be smart enough not to create a duplicate.

wikumChamith: @mogoodrich do you where is the other `org.openmrs.module.emrapi` concept resource is created? One with the description "Reference application module".

dkayiwa: @wikumChamith i think what @mogoodrich meant was that instead of removing the method, just make it smart enough not to create a duplicate. 😊

wikumChamith: @dkayiwa This method already includes a check for duplicates:

https://github.com/openmrs/openmrs-module-emrapi/blob/35032fae6eb22b6be044f8000fcaf4144b846eaf/api/src/main/java/org/openmrs/module/emrapi/EmrApiActivator.java#L349-L350

However, the issue is that the server first runs this method to create the `org.openmrs.module.emrapi` concept source, and then a duplicate is created afterward. So, adding a duplicate check here won’t resolve the problem.

dkayiwa: @wikumChamith can you remove this line and test again? https://github.com/openmrs/openmrs-module-referencemetadata/blob/master/api/src/main/resources/Reference_Application_Concepts-27.xml#L885

wikumChamith: @dkayiwa this is the other duplicate. Is it better to remove `org.openmrs.module.emrapi` from the reference metadata module or remove it from this module?

dkayiwa: Do as i advised above.

wikumChamith: @dkayiwa If we decide to remove it, we'll also need to delete all `concept_reference_terms` that reference it as a foreign key. To do that, we would first need to remove any `concept_reference_maps` that reference those `concept_reference_terms` as foreign keys.

Additionally, the concept source created by this module introduces another issue. Since this module's activator runs before the data from the reference metadata module is loaded, the `org.openmrs.module.emrapi` concept source is added to the database first with a concept_source_id of 1. As a result, [SNOMED CT](https://github.com/openmrs/openmrs-module-referencemetadata/blob/13022e68bb9f6e967bd285df68c3bbfc427bf17f/api/src/main/resources/Reference_Application_Concepts-27.xml#L873) does not get added to the database, and all concept_reference_terms that should reference SNOMED CT are incorrectly mapped to `org.openmrs.module.emrapi`

wikumChamith: It is going to cause a cycle exception.

```
org.openmrs.util.CycleException
[referencemetadata->emrapi, appointmentschedulingui->emrapi, registrationapp->coreapps, appointmentschedulingui->coreapps, emrapi->referencemetadata, attachments->emrapi, htmlformentryui->emrapi, referencedemodata->referencemetadata, coreapps->emrapi, registrationapp->emrapi, referenceapplication->htmlformentryui, referencedemodata->emrapi]
```

dkayiwa: Oh i see!

dkayiwa: @wikumChamith in that case let us just rename this https://github.com/openmrs/openmrs-module-referencemetadata/blob/master/api/src/main/resources/Reference_Application_Concepts-27.xml#L885 to something like `org.openmrs.module.emrapi2`

mogoodrich: Thanks @dkayiwa  @wikumChamith !  

I think I understand what is going, not sure we want to just rename the EMR API concept source in the reference data, let me give this some thought...

mogoodrich: @ibacher  @dkayiwa  what is the history (and ideal future stats?) of the concepts loaded via the reference metadata module vs those loaded via the OCL packages in the reference distro?  I may need to hack around this a bit and I'm wondering if one option may be to install all the EMR API concepts and concept mappings via OCL... (this actually may already be happened).  I'm guessing we are trying to move to install all the concepts via OCL anyway?  fyi @wikumChamith

mseaton: My general take is that we should never have been creating a concept source in the EMR API module activator - as the comment says, this was never supposed to be done in production.  Unfortunately, it's been there since 2013! :)

So I'm in favor of removing this from the activator.  The big question to figure out is how to fix this in systems where it has already been installed.  In these cases, we will need to do an update to replace the uuid of any existing source with the expected uuid as defined elsewhere in that distribution (eg. in MDS packages, etc).

Ideally, core would be smart enough to deal with this, but it doesn't appear to be so.  That would be another place were this could be addressed.  Basically - don't allow creating concept sources with duplicate names.

mogoodrich: Thanks @mseaton !  

Yeah, actually, thinking about what you said, it might be as simple as just removing this and relying on a distro to make sure this mapping is present (if needed).  It looks like both the PIH EMR and the Ref App already do this.

mogoodrich (approved):
So after thinkin through @mseaton 's comments I think he is correct, and we actually can just merge this PR as-is with the original fix from @wikumChamith an no longer create the EMR API concept source.

Any distribution that needs this concept source will need to make sure that it creates it on it's on, but I suspect most/many do... (Ref App and PIH EMR do) fyi @rbuisson 

I don't think we need to worry about changing/cleaning the uuids on this concept source... see my comments above... is there a use case I'm not thinking of @mseaton ?

Approving this PR....

mogoodrich:
Thanks @wikumChamith  and @dkayiwa !

PR #256, opened 2024-12-03
EA-212: Add Support for Fetching first and latest encounter from Inpa…

mogoodrich:
…tient Admission and Inpatient Request


mseaton (approved):
Seems fine to me @mogoodrich , thanks.  Not entirely sure the benefit of the unmodifiableList wrapper, but no opposition to it either...

mogoodrich:
> Seems fine to me @mogoodrich , thanks. Not entirely sure the benefit of the unmodifiableList wrapper, but no opposition to it either...

Yeah, admittedly I was copying some of the stuff you were doing without 100% understanding the benefit/risk.

PR #258, opened 2025-05-06
EA-214 - Make `visitWithDiagnosesAndNotes` endpoint support legacy Di…

mseaton:
…agnosis


api/src/main/java/org/openmrs/module/emrapi/db/EmrApiDAO.java
```diff
@@ -11,15 +8,5 @@ public interface EmrApiDAO {
...
-    /**
-     * Fetches visits with note encounters and diagnoses of a patient.
-     *
-     * @param patient the patient
-     * @param startIndex the start index of the number of visits to fetch
-     * @param limit the limit of the number of visits to fetch
-     *
-     * @return a list of visits that has note encounters and diagnoses
-     */
-    List<VisitWithDiagnosesAndNotes> getVisitsWithNotesAndDiagnosesByPatient(Patient patient, int startIndex, int limit);
```
mseaton, line 24: I had intended to keep this DAO more generic than this.  I have moved thi into the EMRPatientDAO/Impl.

api/src/main/java/org/openmrs/module/emrapi/diagnosis/CoreDiagnosisService.java
```diff
@@ -49,4 +52,22 @@ public List<Obs> getDiagnosesAsObs(Visit visit, DiagnosisMetadata diagnosisMetad
...
+		Map<String, Object> parameters = new HashMap<>();
+		parameters.put("visits", visits);
+		List<org.openmrs.Diagnosis> diagnoses = emrApiDAO.executeHql(query, parameters, org.openmrs.Diagnosis.class);
+		for (org.openmrs.Diagnosis diagnosis : diagnoses) {
+			Visit v = diagnosis.getEncounter().getVisit();
+            List<Diagnosis> l = ret.computeIfAbsent(v, k -> new ArrayList<>());
+            l.add(DiagnosisUtils.convert(diagnosis));
+		}
+		return ret;
+	}
```
mseaton, line 72: This method takes the logic previously implemented within the getVisitsWithNotesAndDiagnosesByPatient DAO, and makes it available within the CoreDiagnosisService so that an alternative can be offered by the ObsGroupDiagnosisService.

_However_, rather than return an `org.openmrs.Diagnosis` object, this returns an `org.openmrs.module.emrapi.diagnosis.Diagnosis` object.   What do we think about this @ibacher / @jayasanka-sack / @mogoodrich ?

api/src/main/java/org/openmrs/module/emrapi/diagnosis/DiagnosisService.java
```diff
@@ -29,6 +30,11 @@ public interface DiagnosisService extends OpenmrsService {
      */
     List<Obs> codeNonCodedDiagnosis(Obs nonCodedObs, List<Diagnosis> diagnoses);
 
+	/**
+	 * @return true if system is configured to store diagnoses as obs groups rather than diagnoses
+	 */
+	boolean useDiagnosesAsObs();
```
mseaton, line 36: This just makes the existing method in the Impl available for use outside of the impl

api/src/main/java/org/openmrs/module/emrapi/diagnosis/DiagnosisServiceImpl.java
```diff
@@ -84,6 +85,15 @@ public List<Obs> codeNonCodedDiagnosis(Obs nonCodedObs, List<Diagnosis> diagnose
...
 	}
 
+	public Map<Visit, List<Diagnosis>> getDiagnoses(List<Visit> visits) {
+		if (useDiagnosesAsObs()) {
+			return obsGroupDiagnosisService.getDiagnoses(visits);
+		}
+		else {
+			return coreDiagnosisService.getDiagnoses(visits);
+		}
+	}
```
mseaton, line 95: This follows the same pattern that we do for diagnoses, where we switch between 2 implementations.

api/src/main/java/org/openmrs/module/emrapi/diagnosis/DiagnosisUtils.java
```diff
@@ -13,16 +13,25 @@ public class DiagnosisUtils {
...
+    public static Diagnosis convert(org.openmrs.Diagnosis coreDiagnosis) {
+        Diagnosis diagnosis = new Diagnosis();
+        CodedOrFreeText coded = coreDiagnosis.getDiagnosis();
+        if (coded != null) {
+            diagnosis.setDiagnosis(new CodedOrFreeTextAnswer(coded.getCoded(), coded.getSpecificName(), coded.getNonCoded()));
+        }
+        diagnosis.setCertainty(coreDiagnosis.getCertainty() == ConditionVerificationStatus.CONFIRMED ? Diagnosis.Certainty.CONFIRMED : Diagnosis.Certainty.PRESUMED);
+        diagnosis.setOrder(coreDiagnosis.getRank() == 1 ? Diagnosis.Order.PRIMARY : Diagnosis.Order.SECONDARY);
+        return diagnosis;
+    }
```
mseaton, line 36: This just pulls the logic in the existing method that converts a List of diagnoses out so that it can be used to convert a single diagnosis.

api/src/main/java/org/openmrs/module/emrapi/diagnosis/ObsGroupDiagnosisService.java
```diff
@@ -143,16 +142,20 @@ public boolean hasDiagnosis(Encounter encounter, Diagnosis diagnosis) {
...
-		List<Obs> observations = obsService.getObservations(Arrays.asList((Person) patient), null, Arrays.asList(diagnosisMetadata.getDiagnosisSetConcept()),
+		List<Obs> observations = obsService.getObservations(Arrays.asList(patient), null, Arrays.asList(diagnosisMetadata.getDiagnosisSetConcept()),
 				null, null, null, Arrays.asList("obsDatetime"),
 				null, null, fromDate, null, false);
+        return getDiagnosesFromObsGroups(observations);
+	}
 
-		for (Obs obs : observations) {
-			Diagnosis diagnosis;
+    protected List<Diagnosis> getDiagnosesFromObsGroups(List<Obs> diagnosisObsGroups) {
```
mseaton, line 152: This just reorganizes the existing code by pulling this out into a method that can be used in multiple places.

api/src/main/java/org/openmrs/module/emrapi/diagnosis/ObsGroupDiagnosisService.java
```diff
@@ -234,15 +227,38 @@ public List<Integer> getAllPatientsWithDiagnosis(DiagnosisMetadata diagnosisMeta
...
+        parameters.put("diagnosisOrderConcept", emrApiProperties.getDiagnosisMetadata().getDiagnosisOrderConcept());
+        List<Obs> obsGroups = emrApiDAO.executeHql(query, parameters, Obs.class);
+        List<Diagnosis> diagnoses = getDiagnosesFromObsGroups(obsGroups);
+        for (Diagnosis diagnosis : diagnoses) {
+            Visit visit = diagnosis.getExistingObs().getEncounter().getVisit();
+            List<Diagnosis> l = ret.computeIfAbsent(visit, k -> new ArrayList<>());
+            l.add(diagnosis);
+        }
+        return ret;
+    }
```
mseaton, line 251: This provides the ObsGroup-based implementation of getting Diagnoses for a list of visits, to implement the crux of this ticket.

api/src/main/java/org/openmrs/module/emrapi/patient/EmrPatientServiceImpl.java
```diff
@@ -82,4 +78,35 @@ public Patient findPatientByPrimaryId(String primaryId) {
...
+		Map<Visit, List<Obs>> notesObs = getVisitNoteObservations(visits);
+		Map<Visit, List<Diagnosis>> diagnoses = emrDiagnosisService.getDiagnoses(visits);
+        return visits.stream().map(visit -> {
+			VisitWithDiagnosesAndNotes note = new VisitWithDiagnosesAndNotes();
+			note.setVisit(visit);
+			note.setDiagnoses(diagnoses.computeIfAbsent(visit, k -> new ArrayList<>()));
+			note.setVisitNotes(notesObs.computeIfAbsent(visit, k -> new ArrayList<>()));
+			return note;
+		}).collect(Collectors.toList());
+	}
```
mseaton, line 111: The goal here was to break down the 3 different queries in the getVisitsWithNotesAndDiagnosesByPatient method into 3 separate methods that are generally available in the EmrPatientService

omod/src/main/java/org/openmrs/module/emrapi/web/controller/VisitController.java
```diff
@@ -30,9 +32,12 @@
 @Setter
 @Controller
 public class VisitController extends BaseRestController {
```
mseaton, line 34: Ultimately, this endpoint is the entire reason for this ticket.  I wonder if we might consider getting rid of VisitWithDiagnosesAndNotes class and the service methods to getVisitsWithDiagnosesAndNotesByPatient, and just turn this into a custom `visit` endpoint that supports additional custom properties that could be expanded in the future as needed.

i.e. have a `visit` endpoint that returns a visit by default, but which - if the user asks for either a particular representation - includes diagnoses and/or notes.  Thoughts @mogoodrich / @ibacher ?

api/src/main/java/org/openmrs/module/emrapi/diagnosis/CoreDiagnosisService.java
mseaton: That's just what the emrapi module does @mogoodrich .  Presumably the emrapi Diagnosis was introduced as a compatibility layer for just this reason?

omod/src/main/java/org/openmrs/module/emrapi/web/controller/VisitController.java
mseaton: Huh @mogoodrich ?  I'm not proposing getting rid of a custom endpoint.  I'm just proposing naming it `...emrapi/.../visit` rather than `...emrapi/.../visitWithDiagnosesAndNotes`, and then having flexibility over what it returns via a represenatation

api/src/main/java/org/openmrs/module/emrapi/diagnosis/CoreDiagnosisService.java
mogoodrich: That may be correct... what are the practical differences here, if any,  in the final JSON representation?

omod/src/main/java/org/openmrs/module/emrapi/web/controller/VisitController.java
chibongho: I've talked to @jayasanka-sack and @ibacher about wanting to have this endpoint be configurable to also fetch other resources, like encounters and obs. Without it, it's hard to have the visit table columns be configurable while adopting this endpoint. We want to maintain the performance optimization, but I think making it more generic makes sense.

api/src/main/java/org/openmrs/module/emrapi/diagnosis/CoreDiagnosisService.java
mseaton: I don't know.  It isn't used anywhere yet, and isn't in any released code, so there are no existing functions to look at.  All unit tests that were in place before continue to pass.

omod/src/main/java/org/openmrs/module/emrapi/web/controller/VisitController.java
mseaton: > ah, so you are not changing what it actually fetches at the DAO, just the rep you convert in into

Maybe it's hard to read in the PR, but no, I have not meaningfully changed the DAO, or at least my intention was not to.  The exact same database queries and logic is in place now as was in place previously.

In actual fact, what I'm proposing will continue to be _more_ efficient than what is in place now, as what I'm proposing is that we only actually fetch "diagnoses" if it is requested in the representation, and we only actually fetch "notes" if it is s requested in the representation.  Otherwise we just fetch Visits.  These were always 3 separate queries that were combined in the existing implementation, and they continue to be 3 separate queries.  What I'm saying is that we should just make this a custom `visit` endpoint which can have emrapi-specific representations or properties for emrapi-specific things like "notes" which is based on an EMRAPI "Visit Note Encounter Type" and "diagnoses" which is based on an EMRAPI supported "Diagnosis" adapter class that allows both core Diagnoses objects or Obs-Group-based diagnoses to be the model behind the scenes.

api/src/main/java/org/openmrs/module/emrapi/diagnosis/CoreDiagnosisService.java
mseaton: So looking at this, the emrapi Diagnosis represents a common mapping layer between the 2 approaches, and implements the common components, which are made up of a Diagnosis (which can have a coded value, a specific coded name, and a free-text non-coded value), a certainty (which can be CONFIRMED or PRESUMED), and an order (which can be PRIMARY or SECONDARY).

The core diagnosis  supports more than this, but EMR API will not map more than this.  Specifically, `Condition` is on the core Diagnosis but is not included in the EMR API version, same with "attributes" and "formNamespaceAndPath".  It also supports an arbitrary "rank" which makes the "order" more flexible.  EMR API only supports 2 "ranks".  A rank of 1 is mapped to PRIMARY, and any other rank is mapped to SECONDARY.

So it is not apples to apples, but likely has the main components that are needed, and we could expand it if need be to better match the core diagnosis.

If we want to create an alternative mapping that would use the core `org.openmrs.Diagnosis` as the common object, we could certainly do that.  The potential problem with that approach is that this represents an actual piece of data (it has primary key and a uuid and represents an entity in the diagnosis table), whereas the emrapi Diagnosis is more of a wrapper of other data to make it more accessible.

mseaton: > FWIW, [DiagnosisUtils.java](https://github.com/openmrs/openmrs-module-emrapi/blob/f03603888100d996ba1b58e4d2706261944e6c48/api/src/main/java/org/openmrs/module/emrapi/diagnosis/DiagnosisUtils.java) has a converter

Exactly, see my comment made above (at the same time you made your comment).  I think we could certainly make a mapper that goes to `org.openmrs.Diagnosis`.  But as I wrote above, I think using the adapter class might be best anyway.

mseaton: @chibongho I think I'm going to quickly spike on doing what you indicated and converting to a core diagnosis and see how that feels to us...

omod/src/main/java/org/openmrs/module/emrapi/web/controller/VisitController.java
mogoodrich: this makes sense, thanks @mseaton ... sorry I got caught up in the explanation without delving into the code itself.

mseaton: OK, I have changed the code around based on this.  It now has the following:

* The endpoint is now a `visit` endpoint.  By default, this returns an array of visit + diagnoses + notes in the given representation, if this representation is not custom.  If custom, it will return the custom rep, with the additional feature that it will only execute the diagnoses and notes queries if they are requested in the representation.  So this should add to the efficiency of the method as well as allow us to add more things in the future that are needed for particular requests without slowing the endpoint down for every existing use case.

* This now converts all Diagnoses into the core Diagnosis, as this allows us to represent everything in both models without any loss of data, and is also future proof so that UIs can be developed against a core Diagnosis interface without needing to know anything about EMR API.

@chibongho / @mogoodrich / @ibacher FYI

Remaining TODO:  Add additional unit tests that actually demonstrate getting diagnoses from Obs using this endpoint...

mogoodrich: That sounds perfect, @mseaton thanks!

PR #259, opened 2025-05-15
EA-215: Create global property to disable visit assignment handler

mogoodrich:


mseaton (approved):
Seems fine to me @mogoodrich , thanks!

PR #263, opened 2025-07-16
EA-219: Update to OpenMRS Platform 2.8.x

wikumChamith:
Ticket: https://openmrs.atlassian.net/browse/EA-219

Legacy branch: https://github.com/openmrs/openmrs-module-emrapi/tree/2.x


wikumChamith:
@dkayiwa, why are we seeing a forbidden class exception even though the MetadataPackagesConfig class is on the whitelisting list?

https://github.com/openmrs/openmrs-module-emrapi/blob/5d0e79d07bc498591360801ffa50ceaeecfacabb/omod/src/main/resources/config.xml#L151-L155

api/src/main/java/org/openmrs/module/emrapi/account/provider/CoreProviderService.java
```diff
@@ -38,11 +39,11 @@ public Provider saveProvider(Provider provider) {
...
+        return provider.getProviderRole();
     }
 
     @Override
-    public void setProviderRole(Provider provider, Object providerRole) {
-        throw new IllegalStateException("Unable to set the provider role, the provider management module is not installed");
+    public void setProviderRole(Provider provider, ProviderRole providerRole) {
+        provider.setProviderRole(providerRole);
     }
 }
```
mseaton, line 49: We should be able to delete this class, no?

api/src/main/java/org/openmrs/module/emrapi/account/provider/ProviderServiceFacade.java
```diff
@@ -18,5 +19,5 @@ public interface ProviderServiceFacade {
 
     Object getProviderRole(Provider provider);
 
-    void setProviderRole(Provider provider, Object providerRole);
+    void setProviderRole(Provider provider, ProviderRole providerRole);
 }
```
mseaton, line 23: We should be able to delete this class, no?

api/src/test/java/org/openmrs/module/emrapi/account/AccountDomainWrapperTest.java
```diff
@@ -45,7 +45,7 @@
 import static org.openmrs.util.OpenmrsConstants.USER_PROPERTY_LOCKOUT_TIMESTAMP;
 import static org.openmrs.util.OpenmrsConstants.USER_PROPERTY_LOGIN_ATTEMPTS;
 
-public class AccountDomainWrapperTest extends EmrApiContextSensitiveTest {
+public class AccountDomainWrapperTest extends BaseModuleContextSensitiveTest {
```
mseaton, line 48: If you are going to switch from EmrApiContextSensitiveTest to BaseModuleContextSensitiveTest, why aren't you deleting the (now empty) EmrApiContextSensitiveTest.  Any reason not to keep it around and use it even if it is empty?  Gives a convenient place to put code that you need to execute for every test in this module.

api/src/test/java/org/openmrs/module/emrapi/account/AccountDomainWrapperTest.java
```diff
@@ -71,28 +71,28 @@ public class AccountDomainWrapperTest extends EmrApiContextSensitiveTest {
...
+    @BeforeEach
     public void setup() {
         accountService = mock(AccountService.class);
         userService = mock(UserService.class);
         personService = mock(PersonService.class);
         providerService = mock(ProviderService.class);
-        providerManagementService = mock(ProviderManagementService.class);
-        providerServiceFacade = new MockProviderServiceFacade(new ProviderManagementProviderService(providerService, providerManagementService));
+        providerManagementService = mock(ProviderService.class);
+        providerServiceFacade = new MockProviderServiceFacade(new CoreProviderService(providerService));
```
mseaton, line 81: This entire providerServiceFacade idea is no longer needed.  Just remove it and use providerService directly

api/src/test/java/org/openmrs/module/emrapi/account/AccountDomainWrapperTest.java
```diff
@@ -566,7 +566,7 @@ public void testShouldNotModifyExistingProviderIdentifier() throws Exception {
         ProviderRole providerRole = new ProviderRole();
 
         providerIdentifierGenerator = mock(ProviderIdentifierGenerator.class);
-        when(providerIdentifierGenerator.generateIdentifier(any(Provider.class))).thenReturn("456");
+        lenient().when(providerIdentifierGenerator.generateIdentifier(any(Provider.class))).thenReturn("456");
```
mseaton, line 569: What do these lenient() calls do?

omod/pom.xml
```diff
@@ -235,18 +235,6 @@
...
 			</plugin>
-			<plugin>
-				<groupId>org.apache.maven.plugins</groupId>
-				<artifactId>maven-surefire-plugin</artifactId>
-				<version>2.6</version>
-				<configuration>
-					<argLine>-Xmx512m -XX:MaxPermSize=512m -Djdk.net.URLClassPath.disableClassPathURLCheck=true</argLine>
-					<includes>
-						<include>**/*Test.java</include>
-						<include>**/*_IT.java</include>
```
mseaton, line 246: Just want to confirm that after removing this that all tests are still executed (eg all _IT tests)

wikumChamith:
Should I remove the MetadataUtilTest, since MetadataUtil has been deprecated and its functionality has been moved to the Metadata Sharing module?

mseaton:
> Should I remove the MetadataUtilTest, since MetadataUtil has been deprecated and its functionality has been moved to the Metadata Sharing module?

I think this is fine @wikumChamith .  If you do so, please also remove MetadataUtil.  It has been deprecated for a long time, and in any event this new change you are introducing definitely requires a new major version anyway.

So please update this PR to change the version in the pom to 3.0.0-SNAPSHOT.

mseaton (approved):
LGTM (though I don't know about the changes to the findPatient DAO, I didn't review these in detail and don't have a good sense of why they are required)

wikumChamith:
> LGTM (though I don't know about the changes to the findPatient DAO, I didn't review these in detail and don't have a good sense of why they are required)

[PatientSearchCriteria](https://github.com/openmrs/openmrs-core/blob/14924a51ab0e0ade83ebfdc2d6e45ca691ca3eb7/api/src/main/java/org/openmrs/api/db/hibernate/PatientSearchCriteria.java#L40-L41) has been deprecated, and this was showing some errors. So, I decided to replace it with the JPA Criteria API, which is the recommended approach starting from Platform 2.7.0.

dkayiwa:
@wikumChamith did you get a chance to look at the build errors?

wikumChamith:
> @wikumChamith did you get a chance to look at the build errors?

@dkayiwa Looks like these errors are coming from core—specifically after this commit:
https://github.com/openmrs/openmrs-core/commit/ec1f42c0e76547c3d72445e7db42d33e8f5218e7

cc: @rkorytkowski

dkayiwa:
@rkorytkowski are we reverting simply for tests? How about making tests hack this out without having to revert the changes in core, if they were for the better running of the application?

wikumChamith:
> The issue is TestingApplicationContext overrides sessionFactory from core and doesn't set cacheConfig property. I think I need to move cacheConfig to be autowired again in case of this kind of overwrite... Overwriting bean definition is hacky and we will want to come up with some better approach here. I'll think about it.

Is this a feature or a bug?

api/src/main/java/org/openmrs/module/emrapi/utils/GeneralUtils.java
```diff
@@ -88,7 +88,7 @@ public static Locale normalizeLocale(String localeString) {
         for (int i = 0; i < len; i++) {
             char c = localeString.charAt(i);
             // allow only ASCII letters and "_" character
-            if ((c <= 0x20 || c >= 0x7f) || ((c >= 0x20 || c <= 0x7f) && (!Character.isLetter(c) && c != 0x5f))) {
```
dkayiwa, line 91: Is there a reason why you dropped `((c >= 0x20 || c <= 0x7f)`?

dkayiwa:
Does this compile on Java 21?

api/src/main/java/org/openmrs/module/emrapi/patient/HibernateEmrPatientDAO.java
```diff
@@ -41,46 +48,44 @@ public class HibernateEmrPatientDAO implements EmrPatientDAO {
...
 		}
-		
-		return (List<Patient>) criteria.list();
+
+		cq.select(patient)
+				.where(cb.and(preds.toArray(new Predicate[0])))
+				.distinct(true);
+
+		TypedQuery<Patient> typed = em.createQuery(cq);
+		if (start != null)     typed.setFirstResult(start);
```
dkayiwa, line 82: https://openmrs.atlassian.net/wiki/spaces/docs/pages/25477044/Java+Conventions#Code-Style

api/src/main/java/org/openmrs/module/emrapi/patient/HibernateEmrPatientDAO.java
```diff
@@ -41,46 +48,44 @@ public class HibernateEmrPatientDAO implements EmrPatientDAO {
...
+		if (maxResults != null) typed.setMaxResults(maxResults);
+
+		return typed.getResultList();
+
 	}
 	
-	private Criteria buildCriteria(String query, Criteria criteria) {
```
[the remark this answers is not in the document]
wikumChamith: PatientSearchCriteria is deprecated. The same logic is now inside the `findPatients` method.
```

if (StringUtils.isNotBlank(query)) {
	if (query.matches(".*\\d.*")) {
		Join<Patient, PatientIdentifier> ids = patient.join("identifiers", JoinType.LEFT);
		preds.add(cb.like(cb.lower(ids.get("identifier")),
				"%" + query.toLowerCase() + "%"));
	} else {
		Join<Patient, PersonName> names = patient.join("names");
		String like = "%" + query.toLowerCase() + "%";
		preds.add(cb.or(cb.like(cb.lower(names.get("givenName")),  like),
				cb.like(cb.lower(names.get("familyName")), like)));
	}
}
```

wikumChamith:
> Does this compile on Java 21?

No. There are some modules that this depends on that do not yet support Java 21.

dkayiwa:
Can't we do as for other modules that we have given support for all the way from Java 8 to 21?

api/src/main/java/org/openmrs/module/emrapi/utils/MetadataUtil.java
```diff
@@ -1,341 +0,0 @@
-/**
- * The contents of this file are subject to the OpenMRS Public License
```
dkayiwa, line 2: @wikumChamith why did we delete this file?

dkayiwa: `org.openmrs.module.ModuleException: Unable to start OpenMRS. Error thrown was: org/openmrs/module/emrapi/utils/MetadataUtil`

wikumChamith: > `org.openmrs.module.ModuleException: Unable to start OpenMRS. Error thrown was: org/openmrs/module/emrapi/utils/MetadataUtil`

In which module?

dkayiwa: When starting O2

PR #270, opened 2026-02-09
O3-5077: Add support for recording Procedures

jayasanka-sack:
https://openmrs.atlassian.net/browse/O3-5077

Talk: https://talk.openmrs.org/t/procedures-app-architecture-questions/47948/

The purpose of this PR is to add support for recording procedures.

It introduces two new entities:

- Procedure
- ProcedureType (e.g., `Historical`, `Current`, etc., with support for custom types)

This PR also includes full CRUD endpoints for both `Procedure` and `ProcedureType`.


dkayiwa:
@jayasanka-sack i guess you are still working on the above build failure. :)

api/src/main/java/org/openmrs/module/emrapi/procedure/ProcedureService.java
```diff
@@ -0,0 +1,51 @@
...
+import org.openmrs.api.APIException;
+import org.openmrs.api.OpenmrsService;
+
+import java.util.List;
+
+/**
+ * Service interface for managing Procedure records.
+ * Provides methods for creating, retrieving, updating, and voiding procedures.
+ */
+public interface ProcedureService extends OpenmrsService {
```
dkayiwa: https://openmrs.atlassian.net/wiki/spaces/docs/pages/25477044/Java+Conventions#New-Public-Methods-and-Classes

omod/src/main/java/org/openmrs/module/emrapi/rest/resource/ProcedureResource.java
```diff
@@ -0,0 +1,131 @@
...
+import org.openmrs.module.webservices.rest.web.representation.Representation;
+import org.openmrs.module.webservices.rest.web.resource.api.PageableResult;
+import org.openmrs.module.webservices.rest.web.resource.impl.DelegatingCrudResource;
+import org.openmrs.module.webservices.rest.web.resource.impl.DelegatingResourceDescription;
+import org.openmrs.module.webservices.rest.web.resource.impl.NeedsPaging;
+import org.openmrs.module.webservices.rest.web.response.ResourceDoesNotSupportOperationException;
+import org.openmrs.module.webservices.rest.web.response.ResponseException;
+
+@Resource(name = RestConstants.VERSION_1 + "/procedure", supportedClass = Procedure.class, supportedOpenmrsVersions = { "2.2 - 9.*" })
+public class ProcedureResource extends DelegatingCrudResource<Procedure> {
```
ibacher: This should actually extend `DataDelegatingCrudResource`

api/src/main/resources/liquibase.xml
```diff
@@ -0,0 +1,266 @@
...
+            <column name="patient_id"/>
+        </createIndex>
+
+        <createIndex indexName="idx_procedure_start_date" tableName="emrapi_procedure">
+            <column name="start_date_time"/>
+        </createIndex>
+
+        <createIndex indexName="idx_procedure_encounter" tableName="emrapi_procedure">
+            <column name="encounter_id"/>
+        </createIndex>
```
ibacher: We may want to double-check that these indexes are correct for the queries we need to run. By this I mean, enabling logging of the Hibernate queries then taking those queries and using them with MySQL `EXPLAIN ...` to see if there are mixed indexes, etc.

api/src/test/resources/ProcedureDataset.xml
```diff
@@ -0,0 +1,137 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<dataset>
+    <!-- Use existing concepts from standard test data -->
+    <!-- Concept 3 and 4 exist in standard test data -->
+
+    <!-- Use existing location from standard test data -->
```
ibacher: These comments are more confusing than elucidating

api/src/main/java/org/openmrs/module/emrapi/procedure/HibernateProcedureDAO.java
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Hibernate implementation of {@link ProcedureDAO}.
+ * @since 3.3.0
+ */
+@Setter
+public class HibernateProcedureDAO implements ProcedureDAO {
+
+   private static final Logger log = LoggerFactory.getLogger(HibernateProcedureDAO.class);
+
+   private DbSessionFactory sessionFactory;
```
ibacher: Since this module depends on Core 2.8.0, it's better to use the `SessionFactory` than the `DbSessionFactory`. See [the docs](https://docs.openmrs.org/doc/org/openmrs/api/db/hibernate/DbSessionFactory.html).

api/src/main/java/org/openmrs/module/emrapi/procedure/ProcedureValidator.java
```diff
@@ -0,0 +1,74 @@
...
+         }
+         if (procedure.getStatus() == null) {
+            errors.reject("Procedure.error.statusRequired");
+         }
+         if (procedure.getVoided() && StringUtils.isBlank(procedure.getVoidReason())) {
+            errors.reject("Procedure.error.voidReasonRequiredWhenVoided");
+         }
+         
+         if (procedure.getProcedureType() != null && procedure.getProcedureType().getUuid().equals(CURRENT_PROCEDURE_TYPE_UUID)){
+            if (procedure.getProcedureType().getRetired()) {
```
ibacher: This rule should probably only be enforced for net-new procedures. It doesn't seem right to enforce this rule when _updating_ an existing procedure.

ibacher (changes requested):
Thanks @jayasanka-sack! Some quick thoughts on this. Mostly looks good to me.

ibacher (changes requested):
Thanks Jaye! This basically looks good to me, so here are some more nit-picking comments. One thing: for Java work on OpenMRS, your IDE should be setup to use tabs, since that's what the OpenMRS Formatter calls for.

api/src/main/java/org/openmrs/module/emrapi/procedure/ProcedureServiceImpl.java
```diff
@@ -0,0 +1,204 @@
...
+	}
+	
+	@Override
+	public Procedure unvoidProcedure(Procedure procedure) {
+		log.info("Unvoiding procedure: {}", procedure.getUuid());
+		
+		procedure.setVoided(false);
+		procedure.setVoidReason(null);
+		procedure.setDateVoided(null);
+		procedure.setVoidedBy(null);
```
jayasanka-sack: already handled by AOP

api/src/test/java/org/openmrs/module/emrapi/procedure/ProcedureDAOTest.java
```diff
@@ -0,0 +1,366 @@
...
+import org.openmrs.Patient;
+import org.openmrs.api.ConceptService;
+import org.openmrs.api.PatientService;
+import org.openmrs.test.jupiter.BaseModuleContextSensitiveTest;
+import org.springframework.beans.factory.annotation.Autowired;
+
+import java.util.Date;
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.*;
```
jayasanka-sack: Remove this.

api/src/main/java/org/openmrs/module/emrapi/procedure/HibernateProcedureDAO.java
```diff
@@ -0,0 +1,166 @@
...
+
+/**
+ * Hibernate implementation of {@link ProcedureDAO}.
+ * @since 3.3.0
+ */
+@Setter
+@Slf4j
+public class HibernateProcedureDAO implements ProcedureDAO {
+	
+	private SessionFactory sessionFactory;
```
wikumChamith: It's better to make this final and use a `@RequiredArgsConstructor` to inject.

api/src/main/java/org/openmrs/module/emrapi/procedure/HibernateProcedureDAO.java
```diff
@@ -0,0 +1,166 @@
...
+		
+		return getEntityManager()
+				.createQuery(criteria)
+				.getSingleResult();
+	}
+	
+	@Override
+	public void deleteProcedure(Procedure procedure) {
+		log.debug("Deleting procedure: {}", procedure.getUuid());
+		sessionFactory.getCurrentSession().delete(procedure);
```
wikumChamith: Why are we mixing JPA and Hibernate Session APIs?

api/src/main/java/org/openmrs/module/emrapi/procedure/Procedure.java
```diff
@@ -0,0 +1,146 @@
...
+	
+	@ManyToOne
+	@JoinColumn(name = "body_site_coded", nullable = false)
+	private Concept bodySite;
+	
+	@Column(name = "start_date_time", nullable = false)
+	private Date startDateTime;
+	
+	@Column(name = "estimated_start_date")
+	private String estimatedStartDate;
```
wikumChamith, line 72: Is there a reason to store this as a `String`?

api/src/main/java/org/openmrs/module/emrapi/procedure/ProcedureService.java
```diff
@@ -0,0 +1,158 @@
...
+	
+	/**
+	 * Saves a procedure record.
+	 * Validates required fields before saving.
+	 *
+	 * @param procedure the procedure to save
+	 * @return the saved procedure
+	 * @throws APIException if validation fails
+	 */
+	Procedure saveProcedure(Procedure procedure) throws APIException;
```
wikumChamith, line 33: Don't we need to introduce some privileges to manage procedures?

api/src/main/resources/liquibase.xml
```diff
@@ -0,0 +1,258 @@
...
+            <column name="estimated_start_date" type="varchar(35)">
+                <constraints nullable="true"/>
+            </column>
+            <column name="end_date_time" type="datetime">
+                <constraints nullable="true"/>
+            </column>
+            <column name="duration" type="int">
+                <constraints nullable="true"/>
+            </column>
+            <column name="duration_unit_coded" type="int">
```
wikumChamith, line 127: Don't we need a foreign key constraint for this?

api/src/main/java/org/openmrs/module/emrapi/procedure/ProcedureValidator.java
```diff
@@ -0,0 +1,73 @@
...
+@Slf4j
+public class ProcedureValidator implements Validator {
+	
+	@Override
+	public boolean supports(Class<?> clazz) {
+		return Procedure.class.isAssignableFrom(clazz);
+	}
+	
+	@Override
+	public void validate(Object target, Errors errors) {
```
wikumChamith, line 26: Don't we need to validate `endDateTime > startDateTime`?

api/src/main/java/org/openmrs/module/emrapi/procedure/ProcedureType.java
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Metadata entity representing a type/category of procedure.
+ * @since 3.3.0
+ */
+@Getter
+@Setter
+@Entity
+@Table(name = "emrapi_procedure_type")
+public class ProcedureType extends BaseOpenmrsMetadata {
```
wikumChamith, line 28: Do we need this? This just extends BaseOpenmrsMetadata with zero additional fields.

omod/src/main/java/org/openmrs/module/emrapi/rest/resource/ProcedureResource.java
```diff
@@ -0,0 +1,152 @@
...
+	
+	@Override
+	public Procedure getByUniqueId(String uuid) {
+		return Context.getService(ProcedureService.class).getProcedureByUuid(uuid);
+	}
+	
+	@Override
+	protected PageableResult doSearch(RequestContext context) throws ResponseException {
+		String patientUuid = context.getParameter("patient");
+		boolean includeVoided = Boolean.parseBoolean(context.getParameter("includeVoided"));
```
wikumChamith: We normally use the `includeAll` parameter.

api/src/main/java/org/openmrs/module/emrapi/procedure/ProcedureServiceImpl.java
```diff
@@ -0,0 +1,204 @@
...
+	@Override
+	@Transactional(readOnly = true)
+	public List<Procedure> getProceduresByPatient(Patient patient, boolean includeVoided, Integer firstResult,
+			Integer maxResults) {
+		log.debug("Getting procedures for patient: {}, includeVoided: {}, firstResult: {}, maxResults: {}", patient,
+				includeVoided, firstResult, maxResults);
+		return procedureDAO.getProceduresByPatient(patient, includeVoided, firstResult, maxResults);
+	}
+	
+	@Override
```
wikumChamith, line 60: Should this be using `@Transactional(readOnly = true)`?

api/src/main/java/org/openmrs/module/emrapi/procedure/HibernateProcedureDAO.java
```diff
@@ -0,0 +1,166 @@
...
+		
+		String jpql = "SELECT pt FROM ProcedureType pt"
+				+ (includeRetired ? "" : " WHERE pt.retired = false")
+				+ " ORDER BY pt.name ASC";
+		
+		return getEntityManager().createQuery(jpql, ProcedureType.class).getResultList();
+	}
+	
+	@Override
+	public ProcedureType saveOrUpdateProcedure(ProcedureType procedureType) {
```
wikumChamith: `saveOrUpdateProcedure` or `saveOrUpdateProcedureType`?

jayasanka-sack: My bad, it should be `saveOrUpdateProcedureType`

api/src/main/java/org/openmrs/module/emrapi/procedure/Procedure.java
jayasanka-sack: Yes, the estimatedStartDate is a partial date, ex: It can be either `2020`, `2020-01`, etc. 
The estimated start date  will be present when creating a historical procedure and it would be used to calculate the startDateTime. 

The estimated date will be still stored to display in the UI when needed and also for future references.

api/src/main/java/org/openmrs/module/emrapi/procedure/ProcedureService.java
jayasanka-sack: Yes, we do. Could you share some references? That would be helpful.

api/src/main/java/org/openmrs/module/emrapi/procedure/ProcedureServiceImpl.java
jayasanka-sack: Good catch!

api/src/main/java/org/openmrs/module/emrapi/procedure/ProcedureType.java
jayasanka-sack: Assuming you ask whether we need the table, if so yes, we need to store the type in a separate table.

api/src/main/java/org/openmrs/module/emrapi/procedure/ProcedureValidator.java
jayasanka-sack: Yeah good catch. The validation should be done in the service in this case as the start Date time is calculated at later stage for historical procedures. I'll get that implemented.

jayasanka-sack (commented):
Thanks for the review @wikumChamith! I'll let you know once I push the changes.

api/src/main/java/org/openmrs/module/emrapi/procedure/ProcedureService.java
wikumChamith: https://github.com/openmrs/openmrs-module-billing/blob/7bf75db366a4f0b02213502475802465a2662de7/api/src/main/java/org/openmrs/module/billing/api/BillService.java#L27

api/src/main/java/org/openmrs/module/emrapi/procedure/ProcedureType.java
wikumChamith: Can't this be a concept?

ibacher: Maybe, but that's inconsistent with everything else in OpenMRS.

api/src/main/java/org/openmrs/module/emrapi/procedure/ProcedureValidator.java
ibacher: It should be safe to do in the validator. The validator is only called for the service's `saveProcedure()` method, which should just persist things to the database.

api/src/main/java/org/openmrs/module/emrapi/procedure/HibernateProcedureDAO.java
```diff
@@ -0,0 +1,166 @@
...
+		return query.getResultList();
+	}
+	
+	@Override
+	public List<ProcedureType> getAllProcedureTypes(boolean includeRetired) {
+		log.debug("Getting all procedure types, includeRetired: {}", includeRetired);
+		
+		String jpql = "SELECT pt FROM ProcedureType pt"
+				+ (includeRetired ? "" : " WHERE pt.retired = false")
+				+ " ORDER BY pt.name ASC";
```
ibacher: This is usually the sort of thing that easiest to keep clean with the `Criteria` API stuff. I hate using string concatenation for this.

api/src/main/java/org/openmrs/module/emrapi/procedure/Procedure.java
```diff
@@ -0,0 +1,146 @@
...
+/**
+ * Entity class representing a surgical or medical procedure performed on a patient.
+ *
+ * @since 3.3.0
+ */
+@Getter
+@Setter
+@Entity
+@Table(name = "emrapi_procedure")
+public class Procedure extends BaseOpenmrsData implements FormRecordable {
```
ibacher: I think you can use `BaseFormRecordableOpenmrsData` no?

denniskigen:
Looks like this one should be closed out too https://github.com/openmrs/openmrs-esm-patient-chart/pull/2880.

omod/src/main/java/org/openmrs/module/emrapi/rest/resource/ProcedureResource.java
jayasanka-sack: Both `includeAll` and `includeVoided` used in different places, but I personally prefer `includeVoided` because it’s more self-explanatory. `includeAll` feels a bit ambiguous.

api/src/main/java/org/openmrs/module/emrapi/procedure/Procedure.java
jayasanka-sack: Yes, much cleaner

api/src/main/resources/liquibase.xml
jayasanka-sack: fixed

omod/src/main/java/org/openmrs/module/emrapi/rest/resource/ProcedureResource.java
wikumChamith: Let's stick to the convention :)

api/src/main/java/org/openmrs/module/emrapi/procedure/ProcedureServiceImpl.java
```diff
@@ -0,0 +1,167 @@
...
+		if (procedure.getEstimatedStartDate() != null) {
+			log.debug("Calculating startDateTime from estimatedStartDate: {}", procedure.getEstimatedStartDate());
+			Date calculatedStartDateTime = ProcedureUtil.getDateTimeFromEstimatedDate(procedure.getEstimatedStartDate());
+			procedure.setStartDateTime(calculatedStartDateTime);
+		}
+		
+		if (procedure.getEndDateTime() != null && procedure.getEndDateTime().before(procedure.getStartDateTime())) {
+			log.warn("End date {} is before start date {}", procedure.getEndDateTime(), procedure.getStartDateTime());
+			throw new APIException("Procedure.error.endDateBeforeStartDate");
+		}
```
wikumChamith: Can we move this to the validator?

wikumChamith (approved):

PR #283, opened 2026-04-13
EA-230: Ensure diagnosis concept mappings exist on startup and handle MissingConceptException gracefully

praisekavuma4:
In EmrApiActivator.java,
called createConceptSource() from started() so the org.openmrs.module.emrapi concept source is always created on module startup if it doesn't already exist.
Added ensureDiagnosisConceptsExist() — a new method called from started() that checks whether all required DiagnosisMetadata concepts exist in the database with their correct mappings. If any are missing, it creates both the concept and its mapping automatically. The required concepts are:

Diagnosis Concept Set
Coded Diagnosis
Non-Coded Diagnosis
Diagnosis Order
Diagnosis Certainty

Also added ensureConceptWithMapping() — a private helper method that handles creation of a single concept and its reference term mapping in a safe, idempotent way (skips silently if the mapping already exists).

In MigrateDiagnosisController.java,
added import org.openmrs.module.emrapi.descriptor.MissingConceptException
Wrapped emrApiProps.getDiagnosisMetadata() in a try/catch block for MissingConceptException. If the exception is thrown, the user is now redirected back to the dashboard with a clear, readable error message instead of a raw Java exception page.

In messages.properties
Added the missing message key emrapi.migrateDiagnosis.missingConcept.error.message with a descriptive error message that guides the user on what went wrong.

**Related.**
https://openmrs.atlassian.net/browse/EA-230

### DIFF
--- a/api/src/main/java/org/openmrs/module/emrapi/EmrApiActivator.java
+++ b/api/src/main/java/org/openmrs/module/emrapi/EmrApiActivator.java
@@ -16,10 +16,16 @@ package org.openmrs.module.emrapi;
 import java.io.File;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Locale;
 
 import org.apache.commons.lang.StringUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.openmrs.Concept;
+import org.openmrs.ConceptMap;
+import org.openmrs.ConceptMapType;
+import org.openmrs.ConceptName;
+import org.openmrs.ConceptReferenceTerm;
 import org.openmrs.ConceptSource;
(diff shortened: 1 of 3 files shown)


praisekavuma4:
@dkayiwa , @jayasanka-sack kindly check on this.

jwnasambu (commented):
@praisekavuma4 Thanks for the progress on this task. Kindly you have an  error caused by Spotless formatting violations.  Your IDE (like IntelliJ, eclipse or VS Code) might be configured to use a slightly different code style than what the OpenMRS project requires e.g. using spaces instead of tabs, or leaving blank lines where it shouldn't. To fix the build  failure run this command in your terminal `mvn spotless:apply`

PR #286, opened 2026-05-06
Make body site field nullable in procedures

jayasanka-sack:
It's not always guaranteed that a procedure will have an associated body site. This PR adds a new migration to drop the NOT NULL constraint on the body_site_coded column, allowing procedures to be recorded without specifying a body site.

ex:

- A general counseling procedure
- Blood transfusion
- Physiotherapy session

### DIFF
--- a/api/src/main/java/org/openmrs/module/emrapi/procedure/Procedure.java
+++ b/api/src/main/java/org/openmrs/module/emrapi/procedure/Procedure.java
@@ -62,7 +62,7 @@ public class Procedure extends BaseFormRecordableOpenmrsData {
 	private String procedureNonCoded;
 	
 	@ManyToOne
-	@JoinColumn(name = "body_site_coded", nullable = false)
+	@JoinColumn(name = "body_site_coded", nullable = true)
 	private Concept bodySite;
 	
 	@Column(name = "start_date_time", nullable = false)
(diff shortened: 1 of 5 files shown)


api/src/main/resources/liquibase.xml
```diff
@@ -259,4 +259,16 @@
             <column name="start_date_time"/>
         </createIndex>
     </changeSet>
+
+    <changeSet id="emrapi-202605061002" author="openmrs">
+        <preConditions onFail="MARK_RAN">
+            <columnExists tableName="emrapi_procedure" columnName="body_site_coded"/>
```
wikumChamith, line 265: The precondition just checks that the column exists, which doesn't really guard anything. The drop is safe to re-run anyway, so either tighten it to check the actual NOT NULL constraint or drop it.

api/src/main/java/org/openmrs/module/emrapi/procedure/Procedure.java
```diff
@@ -62,7 +62,7 @@ public class Procedure extends BaseFormRecordableOpenmrsData {
 	private String procedureNonCoded;
 	
 	@ManyToOne
-	@JoinColumn(name = "body_site_coded", nullable = false)
+	@JoinColumn(name = "body_site_coded", nullable = true)
```
wikumChamith, line 65: nullable = true is the default. So we don't explicitly say that.

wikumChamith (commented):
Some minor nitpicks otherwise LGTM

PR #288, opened 2026-05-11
EA-231: getActiveVisitHelper only needs to fetch active visits for a …

mogoodrich:
…patient


mogoodrich:
Maybe shed-painting, but an inefficiency I noticed while working on another ticket, that I felt comfortable pushing up a PR for after having Claude verify my reasoning and write a bunch of tests.  Still likely worth another pair of eyes. :)

cioan (approved):
LGTM!

mseaton (approved):

mogoodrich:
As luck would have, there was a major reformat of the module that was causing merge constraints so I had Claude re-generate.

PR #291, opened 2026-05-19
Qualify StringToProviderRoleConverter bean name to avoid clash with providermanagement

wikumChamith:
`org.openmrs.module.emrapi.account.StringToProviderRoleConverter` and [org.openmrs.module.providermanagement.converter.StringToProviderRoleConverter](https://github.com/openmrs/openmrs-module-providermanagement/blob/647643dc76047ead89d5a7a9fedfe05b5b87f5ca/api/src/main/java/org/openmrs/module/providermanagement/converter/StringToProviderRoleConverter.java#L27) are both annotated with `@Component` and have the same simple class name. Spring's default `@Component` naming generates the same bean ID for both.

We need this for: https://github.com/mekomsolutions/openmrs-module-initializer/pull/323#pullrequestreview-4314298775

### DIFF
--- a/api/src/main/java/org/openmrs/module/emrapi/account/StringToProviderRoleConverter.java
+++ b/api/src/main/java/org/openmrs/module/emrapi/account/StringToProviderRoleConverter.java
@@ -18,9 +18,9 @@ import org.springframework.core.convert.converter.Converter;
 import org.springframework.stereotype.Component;
 
 /**
- * Converts String to ProviderRole. Ideally this would be in core.
+ * Converts String to ProviderRole. Ideally, this would be in core.
  */
-@Component
+@Component("emrapiStringToProviderRoleConverter")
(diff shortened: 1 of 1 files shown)


mogoodrich:
I believe @mseaton moved this class from Provider Management to EMR API as part of retiring the Provider Management module, so iirc the idea is that if you are running the latest EMR API you do not need (and should not be running) the Provider Management?  That being said, not sure if there are any negative ramifications of this change.

mseaton:
Yeah, we can't assume that this is a harmless change, as there may be other modules in the ecosystem that are declaring an autowired dependency with the specific pre-existing name.  I believe this is mostly an issue within the nuances of initializer due to how classes are loaded across projects and support multiple openmrs and module versions, it isn't something needed in actual practice within emrapi.  So it would be best if we could somehow fix this in initializer.  Like make the master branch of initializer require OpenMRS 2.8.0+, and remove support for anything that pre-dates that.

PR #292, opened 2026-05-22
EA-232: Prevent the EmrApiVisitAssignmentHandler from creating a new …

mogoodrich:
…visit for a patient who already has an active visit at another location on the same date

### DIFF
--- a/api/src/main/java/org/openmrs/module/emrapi/adt/EmrApiVisitAssignmentHandler.java
+++ b/api/src/main/java/org/openmrs/module/emrapi/adt/EmrApiVisitAssignmentHandler.java
@@ -137,10 +137,12 @@ public class EmrApiVisitAssignmentHandler extends BaseEncounterVisitHandler impl
 		// there is no suitable visit so create one if there is a mapping encounter type to the visit type via the Global property
 		if (StringUtils.isNotBlank(administrationService
 		        .getGlobalProperty(EmrApiConstants.GP_VISIT_ASSIGNMENT_HANDLER_ENCOUNTER_TYPE_TO_VISIT_TYPE_MAP))) {
-			if ("false".equalsIgnoreCase(administrationService.getGlobalProperty(
-			    EmrApiConstants.GP_VISIT_ASSIGNMENT_HANDLER_ALLOW_OVERLAPPING_VISITS_AT_ANOTHER_LOCATION))
+			if ("false"
+			        .equalsIgnoreCase(administrationService.getGlobalProperty(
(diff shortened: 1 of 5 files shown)


api/src/main/java/org/openmrs/module/emrapi/adt/EmrApiVisitAssignmentHandler.java
```diff
@@ -136,6 +137,13 @@ public void beforeCreateEncounter(Encounter encounter) {
 		// there is no suitable visit so create one if there is a mapping encounter type to the visit type via the Global property
 		if (StringUtils.isNotBlank(administrationService
 		        .getGlobalProperty(EmrApiConstants.GP_VISIT_ASSIGNMENT_HANDLER_ENCOUNTER_TYPE_TO_VISIT_TYPE_MAP))) {
+			if ("false"
```
mseaton, line 140: Can we do something about this formatting @mogoodrich ?

api/src/main/java/org/openmrs/module/emrapi/adt/EmrApiVisitAssignmentHandler.java
```diff
@@ -175,6 +183,18 @@ protected void updateDateActivatedOfOrdersIfNecessary(Set<Order> orders, Date ex
 		}
 	}
 	
+	private boolean hasActiveVisitAtDatetime(List<Visit> candidates, Date when) {
+		if (candidates == null) {
+			return false;
+		}
+		for (Visit candidate : candidates) {
+			if (candidate.getStopDatetime() == null || !candidate.getStopDatetime().before(when)) {
```
mseaton: Wouldn't really hurt to check the startDatetime here, and would be more accurate given this methods name.

mseaton (approved):
Few minor commnets

api/src/main/java/org/openmrs/module/emrapi/adt/EmrApiVisitAssignmentHandler.java
mogoodrich: you can ask @ibacher :)  the build was failing until I ran the spotless code formatter and changed it to be formatted like this.

api/src/main/java/org/openmrs/module/emrapi/adt/EmrApiVisitAssignmentHandler.java
mogoodrich: Fair... and I did this and things didn't work correctly--we need to check against the entire day (since we end up creating a visit that spans the entire day)

PR #293, opened 2026-08-05
EA-233: Create controller to return dose forms, dose form groupos and…

chibongho:
… routes of administration

See: https://openmrs.atlassian.net/browse/EA-233 and [talk thread](https://talk.openmrs.org/t/proposal-for-linking-dosage-forms-to-routes-of-administration/49252).

Now that we have the relations mappings between dose form, dose form groups and routes of administration in CIEL, the PR implements a controller to return those relationships.

Tested by going to the route (`/openmrs/ws/rest/v1/emrapi/doseFormGroups`) and also by adding different representations (`?v=custom:(doseForms:(doseForm:ref,doseFormGroup:ref)`).

Note:
- I can't find a function in `ConceptService` to return a concept by reference term (`CIEL:123`) AND mapping type (`SAME-AS`), and had to make my own `getConceptBySameAsMapping()` function. Calling `ConceptService.getConceptByMapping(term)` fails because, for some dose form groups, we have multiple mappings for the same term ("SAME-AS" and "Route of administration")
- Other than the representation param (`?v=`), the controller takes no input. Not implemented, but we can technically speed up this controller via caching,
- We use "drug form" / "dose form" (and "drug form group" / "dose form group") somewhat interchangeably. I defaulted to "dose" throughout this PR, but don't have a strong preference.

### DIFF
--- a/api/src/main/java/org/openmrs/module/emrapi/EmrApiConstants.java
+++ b/api/src/main/java/org/openmrs/module/emrapi/EmrApiConstants.java
@@ -116,6 +116,8 @@ public class EmrApiConstants {
 	
 	public static final String NARROWER_THAN_CONCEPT_MAP_TYPE_UUID = "43ac5109-7d8c-11e1-909d-c80aa9edcf4e";
 	
+	public static final String ROUTE_OF_ADMINISTRATION_CONCEPT_MAP_TYPE_UUID = "8ee5b13d-7d8e-11e1-909d-c80aa9edcf4e";
+	
 	public static final String EMR_CONCEPT_SOURCE_NAME = "org.openmrs.module.emrapi";
 	
 	public static final String EMR_CONCEPT_SOURCE_DESCRIPTION = "Source used to tag concepts used in the EMR API module";
--- a/omod/src/main/java/org/openmrs/module/emrapi/web/controller/DoseFormGroupController.java
+++ b/omod/src/main/java/org/openmrs/module/emrapi/web/controller/DoseFormGroupController.java
@@ -1,0 +1,139 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module.emrapi.web.controller;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+import org.openmrs.Concept;
+import org.openmrs.ConceptClass;
+import org.openmrs.ConceptMap;
+import org.openmrs.ConceptMapType;
+import org.openmrs.ConceptReferenceTerm;
+import org.openmrs.api.APIException;
+import org.openmrs.api.ConceptService;
+import org.openmrs.api.context.Context;
+import org.openmrs.module.emrapi.EmrApiConstants;
+import org.openmrs.module.emrapi.web.controller.types.DoseFormGroupRoutes;
+import org.openmrs.module.emrapi.web.controller.types.DoseFormGroupsResponse;
+import org.openmrs.module.emrapi.web.controller.types.DoseFormMembership;
+import org.openmrs.module.webservices.rest.web.ConversionUtil;
+import org.openmrs.module.webservices.rest.web.RequestContext;
+import org.openmrs.module.webservices.rest.web.RestUtil;
+import org.openmrs.module.webservices.rest.web.representation.Representation;
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.ResponseBody;
+
+@Controller
+@RequestMapping(value = "/rest/**/emrapi/doseFormGroups")
+public class DoseFormGroupController {
+
+	private static final String DRUG_FORM_CONCEPT_CLASS_NAME = "Drug form";
+
+	private static final String DOSE_FORM_GROUP_CONCEPT_CLASS_NAME = "Dose Form Group";
+
+	/**
+	 * Returns a {@link DoseFormGroupsResponse}, which contains mappings for: - dose form to its dose
+	 * form group, which is null for a dose form that belongs to no group - dose form group to its
+	 * routes of administration
+	 *
+	 * @param request the current request, used to resolve the requested representation
+	 * @param response the current response
+	 * @return the dose forms and dose form groups, converted to the requested representation
+	 * @throws APIException if a dose form belongs to more than one dose form group
+	 */
+	@RequestMapping(method = RequestMethod.GET)
+	@ResponseBody
+	public Object getDoseFormGroups(HttpServletRequest request, HttpServletResponse response) {
+		RequestContext context = RestUtil.getRequestContext(request, response, Representation.DEFAULT);
+
+		ConceptService cs = Context.getConceptService();
+		ConceptClass doseFormGroupClass = cs.getConceptClassByName(DOSE_FORM_GROUP_CONCEPT_CLASS_NAME);
+		List<Concept> doseFormGroups = cs.getConceptsByClass(doseFormGroupClass);
+
+		List<DoseFormGroupRoutes> doseFormGroupRoutes = new ArrayList<>();
+		// A dose form may belong to at most one dose form group; the concept set membership does not
+		// enforce this, so track the claiming group and fail if a second group claims the same form
+		Map<String, Concept> claimingGroupByDoseFormUuid = new HashMap<>();
+
+		for (Concept doseFormGroup : doseFormGroups) {
+			for (Concept doseForm : doseFormGroup.getSetMembers()) {
+				Concept claimingGroup = claimingGroupByDoseFormUuid.put(doseForm.getUuid(), doseFormGroup);
+				if (claimingGroup != null) {
+					throw new APIException("Dose form '" + displayOf(doseForm) + "' (" + doseForm.getUuid()
+					        + ") belongs to more than one dose form group: '" + displayOf(claimingGroup) + "' and '"
+					        + displayOf(doseFormGroup) + "'. A dose form may belong to at most one group.");
+				}
+			}
+
+			doseFormGroupRoutes.add(new DoseFormGroupRoutes(doseFormGroup, getRoutesOfAdministration(cs, doseFormGroup)));
+		}
+
+		// Every dose form is reported, whether or not a group claims it. The dose forms are enumerated
+		// from their concept class rather than from group membership so that a dose form belonging to no
+		// group is still returned, with a null dose form group.
+		ConceptClass doseFormClass = cs.getConceptClassByName(DRUG_FORM_CONCEPT_CLASS_NAME);
+		List<DoseFormMembership> doseFormMemberships = new ArrayList<>();
+		for (Concept doseForm : cs.getConceptsByClass(doseFormClass)) {
+			doseFormMemberships.add(new DoseFormMembership(doseForm, claimingGroupByDoseFormUuid.get(doseForm.getUuid())));
+		}
+
+		DoseFormGroupsResponse result = new DoseFormGroupsResponse(doseFormMemberships, doseFormGroupRoutes);
+		return ConversionUtil.convertToRepresentation(result, context.getRepresentation());
+	}
+
+	/**
+	 * Returns the routes of administration mapped to the given dose form groupt add.
+	 */
+	private List<Concept> getRoutesOfAdministration(ConceptService cs, Concept doseFormGroup) {
+		List<Concept> routes = new ArrayList<>();
+		for (ConceptMap conceptMap : doseFormGroup.getConceptMappings()) {
+			ConceptMapType mapType = conceptMap.getConceptMapType();
+			if (EmrApiConstants.ROUTE_OF_ADMINISTRATION_CONCEPT_MAP_TYPE_UUID.equals(mapType.getUuid())) {
+				ConceptReferenceTerm term = conceptMap.getConceptReferenceTerm();
+				Concept route = getConceptBySameAsMapping(cs, term);
+				if (route != null) {
+					routes.add(route);
+				}
+			}
+		}
+		return routes;
+	}
+
+	/**
+	 * Retrieves a concept by ConceptReferenceTerm (conceptSource + code) with SAME-AS mapping. This
+	 * function is similar to ConceptService.getConceptByMapping(). However, that function does not
+	 * check for the mapping type (we need the SAME-AS mapping), and throws an exception if the
+	 * ConceptReferenceTerm has multiple mappings, even if they are of different mapping types.
+	 */
+	private Concept getConceptBySameAsMapping(ConceptService cs, ConceptReferenceTerm term) {
+		List<Concept> candidates = cs.getConceptsByMapping(term.getCode(), term.getConceptSource().getName(), false);
+		for (Concept candidate : candidates) {
+			for (ConceptMap candidateMap : candidate.getConceptMappings()) {
+				if (EmrApiConstants.SAME_AS_CONCEPT_MAP_TYPE_UUID.equals(candidateMap.getConceptMapType().getUuid())
+				        && term.getUuid().equals(candidateMap.getConceptReferenceTerm().getUuid())) {
+					return candidate;
+				}
+			}
+		}
+		return null;
+	}
+
+	private String displayOf(Concept concept) {
+		return concept.getName() == null ? null : concept.getName().getName();
+	}
+}
--- a/omod/src/main/java/org/openmrs/module/emrapi/rest/converter/SimpleBeanConverter.java
+++ b/omod/src/main/java/org/openmrs/module/emrapi/rest/converter/SimpleBeanConverter.java
@@ -19,6 +19,9 @@ import org.openmrs.module.emrapi.diagnosis.DiagnosisMetadata;
 import org.openmrs.module.emrapi.disposition.Disposition;
 import org.openmrs.module.emrapi.disposition.DispositionDescriptor;
 import org.openmrs.module.emrapi.disposition.DispositionObs;
+import org.openmrs.module.emrapi.web.controller.types.DoseFormGroupRoutes;
+import org.openmrs.module.emrapi.web.controller.types.DoseFormGroupsResponse;
+import org.openmrs.module.emrapi.web.controller.types.DoseFormMembership;
 import org.openmrs.module.webservices.rest.SimpleObject;
 import org.openmrs.module.webservices.rest.web.ConversionUtil;
 import org.openmrs.module.webservices.rest.web.representation.CustomRepresentation;
@@ -32,7 +35,8 @@ import java.beans.PropertyDescriptor;
 import java.util.Map;
 
 @Handler(supports = { EmrApiProperties.class, DiagnosisMetadata.class, Disposition.class, DispositionObs.class,
-        DispositionDescriptor.class, Diagnosis.class, }, order = 0)
+        DispositionDescriptor.class, Diagnosis.class, DoseFormGroupsResponse.class, DoseFormMembership.class,
+        DoseFormGroupRoutes.class, }, order = 0)
 public class SimpleBeanConverter<T> implements Converter<T> {
 	
 	private final Log log = LogFactory.getLog(getClass());
--- a/omod/src/main/java/org/openmrs/module/emrapi/web/controller/types/DoseFormGroupRoutes.java
+++ b/omod/src/main/java/org/openmrs/module/emrapi/web/controller/types/DoseFormGroupRoutes.java
@@ -1,0 +1,30 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module.emrapi.web.controller.types;
+
+import java.util.List;
+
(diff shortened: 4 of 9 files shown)


omod/src/main/java/org/openmrs/module/emrapi/web/controller/DoseFormGroupController.java
```diff
@@ -0,0 +1,139 @@
...
+		Map<String, Concept> claimingGroupByDoseFormUuid = new HashMap<>();
+
+		for (Concept doseFormGroup : doseFormGroups) {
+			for (Concept doseForm : doseFormGroup.getSetMembers()) {
+				Concept claimingGroup = claimingGroupByDoseFormUuid.put(doseForm.getUuid(), doseFormGroup);
+				if (claimingGroup != null) {
+					throw new APIException("Dose form '" + displayOf(doseForm) + "' (" + doseForm.getUuid()
+					        + ") belongs to more than one dose form group: '" + displayOf(claimingGroup) + "' and '"
+					        + displayOf(doseFormGroup) + "'. A dose form may belong to at most one group.");
+				}
```
dkayiwa: This needs fixing before merge. The design settled on in the talk thread you linked is many-to-many, not 1:1. In [post 23](https://talk.openmrs.org/t/proposal-for-linking-dosage-forms-to-routes-of-administration/49252/23) Burke writes up the CIEL content squad decision as "given a drug's dosage_form, find **all of its dose form groups** (all concepts of class Dose Form Group with the dosage form as a member) and then create a list of all routes mapped from those dose form groups", and in [post 19](https://talk.openmrs.org/t/proposal-for-linking-dosage-forms-to-routes-of-administration/49252/19) Andy gives the concrete case: "oral spray is part of the 'oral spray group' and the 'oral' group". His question about collapsing that to 1:1 was never answered yes.

So the first CIEL release that puts one dose form in two groups breaks this endpoint, with no code change on our side, just an OCL sync. And because the throw happens inside the group loop before anything is assembled, the whole response is lost: every client gets an error instead of the dose forms *and* the routes, not just the one ambiguous form. That also cuts against the fallback ibacher described in [post 1](https://talk.openmrs.org/t/proposal-for-linking-dosage-forms-to-routes-of-administration/49252/1), where missing or odd metadata should degrade rather than fail.

I'd make membership a list and drop the exception, so `DoseFormMembership` carries

```java
private List<Concept> doseFormGroups;
```

and the map here becomes a `Map<String, List<Concept>>`. Clients then union the routes across a form's groups, which is exactly the algorithm in post 23. `shouldFailWhenADoseFormBelongsToMoreThanOneDoseFormGroup` would turn into a test that a form in two groups reports both of them.

omod/src/main/java/org/openmrs/module/emrapi/web/controller/DoseFormGroupController.java
```diff
@@ -0,0 +1,139 @@
...
+
+			doseFormGroupRoutes.add(new DoseFormGroupRoutes(doseFormGroup, getRoutesOfAdministration(cs, doseFormGroup)));
+		}
+
+		// Every dose form is reported, whether or not a group claims it. The dose forms are enumerated
+		// from their concept class rather than from group membership so that a dose form belonging to no
+		// group is still returned, with a null dose form group.
+		ConceptClass doseFormClass = cs.getConceptClassByName(DRUG_FORM_CONCEPT_CLASS_NAME);
+		List<DoseFormMembership> doseFormMemberships = new ArrayList<>();
+		for (Concept doseForm : cs.getConceptsByClass(doseFormClass)) {
```
dkayiwa: `getConceptsByClass` does not filter on `retired`. There is no predicate for it in `HibernateConceptDAO`, on 2.8.0 or on master, so retired concepts come back here and from the group query on line 66. I added a retired dose form and a retired dose form group to your fixture and ran the controller: `Retired Tablet`, `Retired Loose` and the group `Retired Group` all showed up in the response.

That means a client filling a dose form picker from `doseForms` offers retired forms to prescribers. It cannot filter them out either: under `?v=ref`, which is what a picker wants, the payload is `{uuid, display, links}`, as your own representation test pins, so there is no `retired` flag to look at.

The `false` you pass to `getConceptsByMapping` on line 124 already keeps retired route concepts out, so the intent is there; it just isn't applied to the two class queries. Filtering both on `!concept.getRetired()` should be all it takes, and I'd want that in before this merges.

omod/src/main/java/org/openmrs/module/emrapi/web/controller/DoseFormGroupController.java
```diff
@@ -0,0 +1,139 @@
...
+		for (Concept candidate : candidates) {
+			for (ConceptMap candidateMap : candidate.getConceptMappings()) {
+				if (EmrApiConstants.SAME_AS_CONCEPT_MAP_TYPE_UUID.equals(candidateMap.getConceptMapType().getUuid())
+				        && term.getUuid().equals(candidateMap.getConceptReferenceTerm().getUuid())) {
+					return candidate;
+				}
+			}
+		}
+		return null;
+	}
```
dkayiwa: On the note in your description about there being no `ConceptService` method that resolves a reference term by mapping type: this module already has one. `EmrConceptDAO.getConceptsMappedTo(Collection<ConceptMapType>, ConceptReferenceTerm)`, implemented in `api/src/main/java/org/openmrs/module/emrapi/concept/HibernateEmrConceptDAO.java`, does it as a single query keyed on the term you already hold, and `EmrConceptServiceImpl.getConceptsSameOrNarrowerThan` is an existing caller. Not a drop-in, since that service method also matches NARROWER-THAN, but a SAME-AS-only method on `EmrConceptService` over the same DAO call is a few lines.

Not blocking, but two reasons I'd rather see it there. The version here queries by code plus source name, which pulls back every concept mapped to that code under any map type and then walks each candidate's full mappings collection in Java to rediscover the term it started from. And the thread asked for a Java API as well as REST, not just a controller: [post 7](https://talk.openmrs.org/t/proposal-for-linking-dosage-forms-to-routes-of-administration/49252/7), [post 12](https://talk.openmrs.org/t/proposal-for-linking-dosage-forms-to-routes-of-administration/49252/12), and the "new API method" framing in [post 23](https://talk.openmrs.org/t/proposal-for-linking-dosage-forms-to-routes-of-administration/49252/23), with HFE named as the consumer that needs it. Moving the resolution into `EmrConceptService` gets you both, and the `api` module is where the rest of this module keeps logic of this kind.

omod/src/main/java/org/openmrs/module/emrapi/web/controller/DoseFormGroupController.java
```diff
@@ -0,0 +1,139 @@
...
+	private List<Concept> getRoutesOfAdministration(ConceptService cs, Concept doseFormGroup) {
+		List<Concept> routes = new ArrayList<>();
+		for (ConceptMap conceptMap : doseFormGroup.getConceptMappings()) {
+			ConceptMapType mapType = conceptMap.getConceptMapType();
+			if (EmrApiConstants.ROUTE_OF_ADMINISTRATION_CONCEPT_MAP_TYPE_UUID.equals(mapType.getUuid())) {
+				ConceptReferenceTerm term = conceptMap.getConceptReferenceTerm();
+				Concept route = getConceptBySameAsMapping(cs, term);
+				if (route != null) {
+					routes.add(route);
+				}
```
dkayiwa: A route term that fails to resolve is dropped here with no trace, and the class has no logger at all, so "Sublingual disappeared from the route dropdown" leaves nothing on the server to look at.

Same shape one level up: when `getConceptClassByName` returns null, which is what happens on any dictionary that has not imported the new CIEL classes, or if the class name is spelled differently, the endpoint answers 200 with `{"doseForms":[],"doseFormGroups":[]}`. I ran it against a database without the fixture to confirm. Empty is the right fallback per [post 1](https://talk.openmrs.org/t/proposal-for-linking-dosage-forms-to-routes-of-administration/49252/1), but it should not be indistinguishable from a misspelled class name.

A `log.warn` in both places would cover it. Not something I'd hold the PR for.

omod/src/main/java/org/openmrs/module/emrapi/web/controller/DoseFormGroupController.java
```diff
@@ -0,0 +1,139 @@
...
+	 *
+	 * @param request the current request, used to resolve the requested representation
+	 * @param response the current response
+	 * @return the dose forms and dose form groups, converted to the requested representation
+	 * @throws APIException if a dose form belongs to more than one dose form group
+	 */
+	@RequestMapping(method = RequestMethod.GET)
+	@ResponseBody
+	public Object getDoseFormGroups(HttpServletRequest request, HttpServletResponse response) {
+		RequestContext context = RestUtil.getRequestContext(request, response, Representation.DEFAULT);
```
dkayiwa: I would default this to `Representation.REF`, the way `EmrApiConfigurationController` does. On your own fixture (four dose forms, two groups, one route) the default representation serialises to 12,786 characters against 1,514 for `ref`, because every Concept drags its names, descriptions, mappings, answers, setMembers and attributes with it, and CIEL ships around 86 dose forms ([post 27](https://talk.openmrs.org/t/proposal-for-linking-dosage-forms-to-routes-of-administration/49252/27)). That is the cost of every call, and as you note there is no caching. Anyone who wants more can ask for `?v=full` or a custom rep.

Changing the default is free today and a breaking change once clients depend on it.

omod/src/main/java/org/openmrs/module/emrapi/web/controller/DoseFormGroupController.java
```diff
@@ -0,0 +1,139 @@
...
+		for (Concept doseForm : cs.getConceptsByClass(doseFormClass)) {
+			doseFormMemberships.add(new DoseFormMembership(doseForm, claimingGroupByDoseFormUuid.get(doseForm.getUuid())));
+		}
+
+		DoseFormGroupsResponse result = new DoseFormGroupsResponse(doseFormMemberships, doseFormGroupRoutes);
+		return ConversionUtil.convertToRepresentation(result, context.getRepresentation());
+	}
+
+	/**
+	 * Returns the routes of administration mapped to the given dose form groupt add.
```
dkayiwa: Stray text at the end of this line: "dose form groupt add". Same kind of thing in the PR title, `groupos`, which is worth fixing since it becomes the commit message on squash.

mseaton (commented):
Will review further one the dust has settled with Daniel's / AI's review comments.  One thing I'd say is that a lot of this is actually more API / Service-level functionality that the web / REST layer should just be exposing in a certain way.

chibongho:
Updated the PR:
- does form maps to a list of (instead of just one) dose form groups 
- `getConceptsByClassName()` moved to `EmrConceptService`, and filters out retired concepts.
- Changed the rep default to `Representation.REF`
- Moved most of the logic to EmrConceptService so that it's usable at the service level

api/src/main/java/org/openmrs/module/emrapi/concept/EmrConceptServiceImpl.java
```diff
@@ -106,4 +120,132 @@ public List<ConceptSearchResult> conceptSearch(String query, Locale locale, Coll
...
+	}
+	
+	@Override
+	@Transactional(readOnly = true)
+	public List<Concept> getRoutesOfAdministration(Concept doseForm) {
+		if (doseForm == null) {
+			throw new IllegalArgumentException("doseForm is required");
+		}
+		
+		List<Concept> doseFormGroups = getDoseFormGroupsByDoseFormUuid().get(doseForm.getUuid());
```
dkayiwa: Worth swapping this for `ConceptService.getSetsContainingConcept(doseForm)`, with the parents filtered down to non-retired concepts of the `Dose Form Group` class. Building the whole group-to-form map is right in `getDoseFormMemberships()`, which reads all of it, but here it gets built so that one key can be read back out, so answering "what routes does Tablet have" walks every dose form group and hydrates every dose form in every group.

I ran both against a fixture sized like CIEL (15 dose form groups, 90 dose forms, 20 routes, the target form in two of the groups). Same six routes come back in the same order, and your 14 tests stay green. The cold call drops from 32 SQL statements, 259 entity loads and 69 collection loads to 25, 29 and 10. A repeat call in the same session is 14 statements either way, so the win is the first call plus the scaling: the current cost tracks the size of the dictionary rather than the size of the answer, and [post 27](https://talk.openmrs.org/t/proposal-for-linking-dosage-forms-to-routes-of-administration/49252/27) already lists around 90 dose forms.

Non-blocking, and I would leave `getDoseFormMemberships()` on the shared helper.

api/src/test/java/org/openmrs/module/emrapi/concept/EmrConceptServiceComponentTest.java
```diff
@@ -83,6 +87,139 @@ public void testGetConceptsSameOrNarrowerThanTerm() throws Exception {
...
+	 * a caller holding a drug's dosage form actually wants to know.
+	 */
+	@Test
+	public void testGetRoutesOfAdministration() throws Exception {
+		executeDataSet("doseFormGroupDataset.xml");
+		executeDataSet("doseFormGroupMultipleGroupsDataset.xml");
+		
+		// Tablet is in Oral Solid, whose route is Oral, and in Oral Liquid, whose routes are Oral and Nasal
+		assertThat(namesOf(emrConceptService.getRoutesOfAdministration(conceptService.getConcept(5011))),
+		    is(setOf("Oral", "Nasal")));
```
dkayiwa, line 203: `namesOf()` collapses the list into a `Set`, so this cannot see a duplicate, and the "distinct routes" guarantee in the `getRoutesOfAdministration` javadoc has nothing holding it up. I deleted the `!routes.contains(route)` guard from the implementation and all 14 tests still passed, with the method returning `[Oral, Nasal, Oral]`. The fixture sets the case up properly already, since Tablet is in two groups that both map to Oral; it is only the `Set` that hides it.

Checking the size of the returned list first covers it. I ran this both ways: it passes as written and fails once the guard is removed.

```suggestion
		// Tablet is in Oral Solid, whose route is Oral, and in Oral Liquid, whose routes are Oral and Nasal
		List<Concept> tabletRoutes = emrConceptService.getRoutesOfAdministration(conceptService.getConcept(5011));
		assertThat("Oral is a route of both of Tablet's groups, and should be reported once", tabletRoutes.size(), is(2));
		assertThat(namesOf(tabletRoutes), is(setOf("Oral", "Nasal")));
```

Worth doing while the test is fresh, though it does not hold anything up.

api/src/main/java/org/openmrs/module/emrapi/concept/EmrConceptServiceImpl.java
chibongho: Good catch. Updated `getRoutesOfAdministration()` to be more efficient by using `getSetsContainingConcept()`. (In practice, a dose form belongs to very few dose form groups so this is more efficient)

api/src/main/java/org/openmrs/module/emrapi/concept/EmrConceptServiceImpl.java
```diff
@@ -106,4 +121,136 @@ public List<ConceptSearchResult> conceptSearch(String query, Locale locale, Coll
...
+		}
+		return memberships;
+	}
+	
+	@Override
+	@Transactional(readOnly = true)
+	public List<DoseFormGroupRoutes> getDoseFormGroupRoutes() {
+		List<DoseFormGroupRoutes> doseFormGroupRoutes = new ArrayList<DoseFormGroupRoutes>();
+		for (Concept doseFormGroup : getConceptsByClassName(EmrApiConstants.DOSE_FORM_GROUP_CONCEPT_CLASS_NAME)) {
+			doseFormGroupRoutes.add(new DoseFormGroupRoutes(doseFormGroup, getRoutesOfDoseFormGroup(doseFormGroup)));
```
dkayiwa, line 145: Each route reference term is resolved once per group that maps to it, rather than once per distinct term, and each resolution is two queries: `getConceptsSameAs` re-fetches the SAME-AS map type through `EmrApiProperties.getSameAsConceptMapType()`, then queries the concepts on the term.

I built a fixture at the scale [post 27](https://talk.openmrs.org/t/proposal-for-linking-dosage-forms-to-routes-of-administration/49252/27) describes: 86 dose forms in 15 dose form groups, 20 routes, four route mappings per group. One call to the endpoint runs 126 queries, 122 of them from this method, and 120 of those are the 60 route lookups. Threading a `Map<ConceptReferenceTerm, Concept>` from here down through `getRoutesOfDoseFormGroup` and checking it before `getRouteConcept` brings that to 46, since only 20 distinct terms are involved. Your 14 tests pass either way.

Not blocking, and it is a different thing from the caching you mention in the description: this one lives and dies inside a single call, so there is nothing to invalidate.

api/src/main/java/org/openmrs/module/emrapi/concept/EmrConceptServiceImpl.java
dkayiwa: The switch to `getSetsContainingConcept()` dropped the non-retired filter on the parents, and that part needs fixing before this merges. `HibernateConceptDAO.getSetsContainingConcept` has no retired predicate on 2.8.0 or on master, it is just `where conceptSet = :concept`, so the class-name check here is the only thing filtering the parents.

`Retired Group` in your fixture has no route mappings, which is why nothing failed. I gave it one and ran it: `getRoutesOfAdministration(Tablet)` comes back `[Oral, Rectal]`, and `Rectal` is reachable only through the retired group. On `f8f90274` the same fixture gave `[Oral]`.

The two halves of the feature now disagree. `getDoseFormMemberships()` reports Tablet's groups as `[Oral Solid]` and `getDoseFormGroupRoutes()` leaves the retired group out, so a REST client unioning routes across a form's groups, which is what the controller javadoc tells it to do, gets `[Oral]` while a Java caller gets `[Oral, Rectal]`. HFE is the named consumer of the Java method ([post 12](https://talk.openmrs.org/t/proposal-for-linking-dosage-forms-to-routes-of-administration/49252/12), [post 23](https://talk.openmrs.org/t/proposal-for-linking-dosage-forms-to-routes-of-administration/49252/23)), so that is where a prescriber gets offered a route CIEL has withdrawn, with nothing thrown and nothing logged.

One line covers it, and all 14 tests stay green:

```java
if (!parentSetConcept.getRetired()
        && EmrApiConstants.DOSE_FORM_GROUP_CONCEPT_CLASS_NAME.equals(parentSetConcept.getConceptClass().getName())) {
```

Worth giving `Retired Group` a route in `doseFormGroupDataset.xml` while you are in there, so `testGetRoutesOfAdministration` actually covers this. Without one the fixture cannot see the difference.

dkayiwa (approved):

api/src/main/java/org/openmrs/module/emrapi/concept/EmrConceptService.java
```diff
@@ -30,17 +30,53 @@ public interface EmrConceptService {
...
 	List<Concept> getConceptsSameOrNarrowerThan(ConceptReferenceTerm term);
 	
+	/**
+	 * Unlike ConceptService.getConceptByMapping(), this does not fail when a term carries mappings of
+	 * more than one type, and it only considers SAME-AS mappings.
+	 *
+	 * @param term
+	 * @return all concepts with SAME-AS mappings to term, including retired ones
+	 */
+	List<Concept> getConceptsSameAs(ConceptReferenceTerm term);
```
mseaton, line 40: Not sure what the use case is for when someone has a ConceptReferenceTerm already and wants to call this - I would have expected we want to take in a source (name or hl7 code string) and term (string), and limit this to only same-as mappings.  Starting with a term means 2 API calls I would think.

api/src/main/java/org/openmrs/module/emrapi/concept/EmrConceptServiceImpl.java
```diff
@@ -106,4 +121,140 @@ public List<ConceptSearchResult> conceptSearch(String query, Locale locale, Coll
...
+		
+		// ConceptService.getSetsContainingConcept() does not filter on retired, so the parents are
+		// filtered here, the same way getConceptsByClassName() filters the other two lookups. A route
+		// reachable only through a retired dose form group has been withdrawn and must not be offered.
+		List<ConceptSet> setsContainingDoseForm = conceptService.getSetsContainingConcept(doseForm);
+		List<Concept> doseFormGroups = new ArrayList<>();
+		for (ConceptSet parentSet : setsContainingDoseForm) {
+			Concept parentSetConcept = parentSet.getConceptSet();
+			if (!parentSetConcept.getRetired()
+			        && EmrApiConstants.DOSE_FORM_GROUP_CONCEPT_CLASS_NAME.equals(parentSetConcept.getConceptClass().getName())) {
```
mseaton, line 165: At bit strange to compare class names as strings here, vs. getting the concept class (by name now, could be by some other mechanism later) and comparing against it.

api/src/main/java/org/openmrs/module/emrapi/concept/EmrConceptServiceImpl.java
```diff
@@ -106,4 +121,140 @@ public List<ConceptSearchResult> conceptSearch(String query, Locale locale, Coll
...
+	}
+	
+	/**
+	 * The routes of administration a dose form group is mapped to. A route reference term that no
+	 * non-retired concept claims via a SAME-AS mapping is skipped rather than failing the lookup.
+	 */
+	private List<Concept> getRoutesOfDoseFormGroup(Concept doseFormGroup) {
+		List<Concept> routes = new ArrayList<Concept>();
+		for (ConceptMap conceptMap : doseFormGroup.getConceptMappings()) {
+			if (EmrApiConstants.ROUTE_OF_ADMINISTRATION_CONCEPT_MAP_TYPE_UUID
```
mseaton, line 208: Same here.  I guess this saves a single db call, but feels better to me to get the map type and compare against it vs. comparing the uuid.  It's fine, and what you have is probably better / more efficient, just doesn't feel totally right to me

api/src/main/java/org/openmrs/module/emrapi/EmrApiConstants.java
```diff
@@ -116,6 +116,16 @@ public class EmrApiConstants {
 	
 	public static final String NARROWER_THAN_CONCEPT_MAP_TYPE_UUID = "43ac5109-7d8c-11e1-909d-c80aa9edcf4e";
 	
+	public static final String ROUTE_OF_ADMINISTRATION_CONCEPT_MAP_TYPE_UUID = "8ee5b13d-7d8e-11e1-909d-c80aa9edcf4e";
```
mseaton, line 119: Can we say definitively that this will always be true?

mseaton (approved):
Seems ok to me generally - I did make a handful of comments, nothing critical though

api/src/main/java/org/openmrs/module/emrapi/EmrApiConstants.java
ibacher: It's part of the OpenMRS [core data](https://github.com/openmrs/openmrs-core/blob/2ce847744295132bbe5fb97f369fab714f09b309/api/src/main/resources/org/openmrs/liquibase/snapshots/core-data/liquibase-core-data-2.5.x.xml#L1259), so I think we can guarantee this one.

chibongho:
Made another update. Added functions to `EmrApiProperties` to return particular `ConceptMapType` (route of administration) and `ConceptClass` (dose form and dose form group), and used them in `EmrConceptService`.

